Issue #213, opened 2022-06-14
[FEAT] - Allow uploading different hero images for different screens (desktop/tablet/phone)


bskinn:
### What site is this for?

Quansight LLC & Labs

### Actual behavior

It is unclear how we will be able to achieve the desired appearance of the hero images across the website as it is currently implemented, where only a single hero image can be uploaded for each page.

To the best of our knowledge, standard practice is to support upload of three hero images for each page, one each for desktop, tablet and mobile views.

Barring detailed guidance on how to achieve the desired design effect with the current Storyblok configuration, we will need this feature added prior to launch.

gabalafou:
I might be misunderstanding this issue. Are you thinking about uploading images with different aspect ratios or different resolutions? Because I think that for different resolutions (for different screen widths), this is already taken care of by the platform. 

You can test it out. First narrow your browser window. Then go to [any page](https://quansight-consulting.vercel.app/library/blog/a-year-in-review-quansight-s-contributions-to-pytorch-in-2021) on the website. Then open Chrome Dev Tools, go to the network panel, clear the accumulated requests, then widen the window while keeping an eye on the network panel. You should notice that the page requests larger image versions as the window gets wider.

gabalafou:
Is this Naturaily or Quansight? I have some thoughts. 

As far as I can tell, the ability to upload images with different aspect ratios was never discussed with Naturaily, nor covered explicitly in [the Figma designs](https://www.figma.com/file/vyWM9qXdXCSAPlTfUjuqrL/Quansight-2021-Website?node-id=1181%3A340) or [the requirements doc](https://docs.google.com/document/d/196XZSMPNzT6-J3DtbI_uHG66-cBk9grh5e0u6_LDt1k/edit).

That said, the mobile screenshots currently do not match the mobile wireframes in Figma, so that part is on Naturaily. I'm not 100% sure but I think all it requires is changing some CSS rules. The Figma files for LLC are not explicit about whether the hero images change dimension on desktop vs mobile, or if they are just "framed" differently.

My understanding is that Irina recently created a set of illustrations for **Labs** that have different dimensions depending on device (desktop vs mobile). However, those were missing from Figma for the past three+ months of development. 

If we absolutely want different aspect ratios for more fine-grained control over how these images get zoomed and/or cropped on portrait versus landscape orientations, then this will require more discussion.

Issue #177, opened 2022-06-10
[INFO] - Specifications for Hero images (size? different versions for mobile/desktop?)


bskinn:
### What site is this for?

Quansight LLC

### Actual behavior

Naturaily, please provide any guidance you have about optimal sizes for Hero images. 

Also, do we need to upload different Hero image versions for desktop vs mobile? If so, how do we indicate these multiple images on the Storyblok Hero component?

Thanks!

bskinn:
This is downstream of #213

gabalafou:
To me, this feels like something missing from [the Storyblok manual](https://docs.google.com/document/d/1HVaWKhepSxIQP1ZY9o7O_n3qCQpE8KdgD9KIaS4P-n4/edit#heading=h.enx0tixw76df).

gabalafou:
at-team-Naturaily: my guess is that we need to upload images at a large width (like 3000 px?) and then the image resizer service will size the image down to smaller widths for screens with lower resolutions. 

Whatever the case, we would love for the exact details to be in the manual.

PR #232, opened 2022-06-19
fix: Hero image should cover the space not shrink to fit

gabalafou:
This could potentially be a fix for #213, at least a partial fix or a stopgap fix.

Previously, the code was supposedly defaulting to `object-fit: contain` but `cover` matches the [Figma designs](https://www.figma.com/file/vyWM9qXdXCSAPlTfUjuqrL/Quansight-2021-Website?node-id=1181%3A340) much better.

Here are two screenshots showing the difference. Both are at mobile device width.

When object-fit is set to `contain`, the hero image is shrunk down so that it fits, which is not desired:

When object-fit is set to `cover`, the hero image expands across the available space, as desired:

I also had to change the code to not use a default function parameter value for `objectFit` because while debugging, I discovered that the CMS was passing an empty string for `objectFit` when a value for object fit is not chosen in the Storyblok UI.

Here's a snippet of the json returned from `http://localhost:4200/_next/data/development/library.json` in which `objectFit` is assigned an empty string:

```json
{
    "pageProps": {
        "data": {
            "__typename": "PageItem",
            "alternates": [],
            "content": {
                "__typename": "PageComponent",
                "_editable": "<!--#storyblok#{\"name\": \"page\", \"space\": \"147759\", \"uid\": \"52531dbb-5aff-4f6f-a084-fcc2da90c1c1\", \"id\": \"143249754\"}-->",
                "_uid": "52531dbb-5aff-4f6f-a084-fcc2da90c1c1",
                "body": [
                    {
                        "_uid": "5432d60a-c2c8-41a9-9131-13d29e123d97",
                        "image": {
                            "id": 4981932,
                            "alt": "Hero decoration",
                            "name": "",
                            "focus": null,
                            "title": "",
                            "filename": "https://a.storyblok.com/f/147759/1440x746/5cfdc24cd0/paris_v6.png",
                            "copyright": "",
                            "fieldtype": "asset"
                        },
                        "title": "Blog",
                        "variant": "medium",
                        "subTitle": "",
                        "component": "hero",
                        "objectFit": "",
                        "_editable": "<!--#storyblok#{\"name\": \"hero\", \"space\": \"147759\", \"uid\": \"5432d60a-c2c8-41a9-9131-13d29e123d97\", \"id\": \"143249754\"}-->"
                    },
```

And here's a screenshot showing the corresponding empty `Object Fit` field in Storyblok:

### DIFF
--- a/libs/shared/ui-components/src/Hero/Hero.tsx
+++ b/libs/shared/ui-components/src/Hero/Hero.tsx
@@ -12,7 +12,7 @@ export const Hero: FC<THeroProps> = ({
   imageSrc,
   imageAlt,
   backgroundColor,
-  objectFit = 'contain',
+  objectFit,
 }) => {
   const isLargeHero =
     variant === HeroVariant.Large || variant === HeroVariant.LargeOverlapping;
@@ -39,7 +39,7 @@ export const Hero: FC<THeroProps> = ({
           src={imageSrc}
           alt={imageAlt}
           layout="fill"
-          objectFit={objectFit}
+          objectFit={objectFit || 'cover'}
           objectPosition="center"
         />
         {title && (


bskinn:
@irinafumarel Is this a valid fix from a design perspective?

If I understood you correctly when we discussed these on Meet last week, I don't think this is consistent with Pierre's vision for the heroes, as data visualizations.

gabalafou:
@bskinn - to be clear, Pierre's vision for the heroes, if you're referring to animated background images, should be an entirely separate ticket from #213, which as I understand it, is only about the ratio and sizing of **static** hero images.

bskinn:
No, agreed, we've abandoned animated hero images.

I believe Pierre does not want the data-containing portions of any of the hero images to be cropped away on any screen size. Irina can confirm/clarify.

gabalafou:
> I believe Pierre does not want the data-containing portions of any of the hero images to be cropped away.

Ah I see. Okay, well just to be clear, that's not at all what was communicated on the Figma boards. For example, here is a screenshot of one of the Figma designs where you can see that the dataviz is cropped at mobile:



I think there have maybe been two competing visions for these data visualizations. 

For Troy, I think they were like art or decoration, in which case you wouldn't worry much about cropping the image. Let's be clear, these are data visualizations, so there is no way to crop them without removing data-containing portions of the image; the entire image contains data. But when it serves as decorative art, it should be okay to crop, since you're less concerned about the data story behind the image and more about the aesthetic value it provides.

For Pierre and others, I think the vision was that the data visualization would draw some attention to itself (through some kind of animated transition) and additionally present some kind of link to a page or a pop-up that explains the visualization and the story that it tells.

I am a huge fan of the second vision, but there has been little concrete work done for it beyond the visualizations themselves, so in my view, this more expansive vision requires a separate project of its own, with allocation of time and resources that go beyond our initial scope of work.

irinafumarel:
@gabalafou @bskinn The data viz was cropped for mobile to illustrate the sizes of the mobile hero so we can upload the mobile hero to Storyblock.  At that time I was communicated that Naturaily will be customizing and animating the heroes so me and Pierre stopped working on this. I will work this week with Pierre to resize that dataviz for mobile.

bskinn:
@gabalafou Agree with everything you wrote here. 

For launch, I think we have to find a solution using static hero images that provides a minimum-attractive solution on all screens (good resolution, no aspect ratio distortion, and good visual layout (no excess padding, etc.)). This may mean we have little choice but to crop the heroes for tablet/mobile, and then figure out a better solution post-launch that avoids cropping.

To be clear, it's entirely possible that existing functionality is sufficient to do this, and we just haven't figured out the right combination of options on the Hero component to achieve it.

To summarize, I think we're looking at a three-stage progression:

1. For launch: A minimum viable product (MVP) Hero solution with static images.
2. Post-launch: If this MVP requires cropping Hero images, developing a solution that does not require cropping of the static images.
3. Post-launch: Developing toward the full "second vision", with animation/transitions and pop-ups leading to explainer pages about each visualization. (One uncertainty here: is this vision compatible with a11y constraints?)

(This might be complex enough to warrant its own Github project/board?)

Comments please!

Crossref #177 and #213.

gabalafou:
No comments to add, just 👍

marcin-antczak-naturaily (approved):

Issue #156, opened 2022-06-03
[a11y] - Labs (staging)


trallard:
- 🔗  [Reference document](https://docs.google.com/document/d/1B3Wr17-1Ic1X6uJe7UQQTzBnfB09rPTatEsax82jIcs/edit)
- [Preview (staging) for LLC](https://github.com/Quansight/Quansight-website/deployments/activity_log?environment=Preview+%E2%80%93+quansight-labs)
- [Production for LLC](https://github.com/Quansight/Quansight-website/deployments/activity_log?environment=Production+%E2%80%93+quansight-consulting](https://github.com/Quansight/Quansight-website/deployments/activity_log?environment=Production+%E2%80%93+quansight-labs))
- 📖  [Gov uk a11y checklist](https://www.gov.uk/service-manual/helping-people-to-use-your-service/understanding-wcag )

---

## Criteria

- [ ] Site is keyboard accessible
	- [ ] Tab order is logical
	- [ ] Focus is never trapped in a loop

- [ ] Forms inputs have explicit labels 
	- [ ] Forms can be filled with just the keyboard
	- [ ] Forms contain title attributes AND/OR
	- [ ] Forms contain legends, labels, hin text, and tooltips
	- [ ] Form instructions are associated with inputs

- [ ] Provide alternative content for non-text content
	- [ ] All images have an alt attribute

- [ ] Text has sufficient color contrast
	- [ ] All text has a contrast ratio of 4.5:1 with the background
	- [ ] Links can be easily distinguished from surrounding text

- [ ]  Headers are nested properly
	- [ ] Only one H1 per page
	- [ ] Headings are nested in a logical way
	- [ ] Subsection hierarchy is preserved

- [ ] Language is set
	- [ ] The language attribute is set in all pages (in this case English throughout)
	
- [ ] Site can be used without stylesheets
	- [ ] The site does not rely on CSS to be functional or understandable
	
- [ ] Links are unique and descriptive
	- [ ] All links can be understood taken alone, e.g., ‘Read more about 508’ OR
	- [ ] Links provide an adequate title or aria attribute
	- [ ] Links that open in a new window indicate so through an aria-label or title

- [ ] Page titles are descriptive
	- [ ] Page titles are in plain English
	- [ ] Titles describe the site and page being displayed

gabalafou:
[...]

Issue #153, opened 2022-06-03
[QA] - Labs site (staging)


trallard:
Please start with #152 

## URLs

- [Preview (staging) for LLC](https://github.com/Quansight/Quansight-website/deployments/activity_log?environment=Preview+%E2%80%93+quansight-labs)
- [Production for LLC](https://github.com/Quansight/Quansight-website/deployments/activity_log?environment=Production+%E2%80%93+quansight-consulting](https://github.com/Quansight/Quansight-website/deployments/activity_log?environment=Production+%E2%80%93+quansight-labs))

### Functional

- [ ] Check [against initial specification doc](https://docs.google.com/document/d/196XZSMPNzT6-J3DtbI_uHG66-cBk9grh5e0u6_LDt1k/edit)
- [ ] Forms and integrations 
	- [ ] Greenhouse (displaying current jobs)
	- [ ] Pipedrive (can use the form and the validation is working - no required fields can be left empty, email validation)
- [ ] Check hyperlinks (wonder if we should look into adding tests for this - through CI - though this might change the workflow with Vercel 🤔 )
	- [ ] Internal hyperlinks
	- [ ] External
- [ ] Check [loading speed](https://pagespeed.web.dev/)
- [ ] Components work as expected (not adding a comprehensive list here - carousels, links, buttons, dropdowns, etc.)

### Browsers

- [ ] Browser compatibility
	- [ ] Mobile
	- [ ] Chromium based web browsers
	- [ ] Safari
	- [ ] Firefox
	- [ ] Edge (?)

### CMS 

- [ ] Preview is working as expected
- [ ] Staging is content is updated once `Publish` in Storyblok is used
- [ ] Metadata for images, content types and others is adequately pre-defined (i.e. fields needed)

### Repository

- [ ] Technical documentation is included
	- [ ] Deployment and maintenance related
	- [ ] Dependencies and libraries are listed
	- [ ] How the repo is structured (where things live)
	- [ ] Coding standards are documented

	
### SEO

trallard:
@gabalafou can you please prioritise this and the accessibility checks for Labs? 
These need to be completed soon and I anticipate will uncover bugs or the such which will need to be added to the queue. Right now all the issues come from the Storyblok use cases and workflows.

gabalafou:
This is indeed my highest priority now that I've helped sort through the GitHub issues. 

I've been working assiduously on the sections of this task "CMS" and "Repository":

[...]

Issue #152, opened 2022-06-03
[QA] - Quansight LLC (staging)


trallard:
## URLs

- [Preview (staging) for LLC](https://github.com/Quansight/Quansight-website/deployments/activity_log?environment=Preview+%E2%80%93+quansight-consulting)
- [Production for LLC](https://github.com/Quansight/Quansight-website/deployments/activity_log?environment=Production+%E2%80%93+quansight-consulting)

### Functional

- [ ] Check [against initial specification doc](https://docs.google.com/document/d/196XZSMPNzT6-J3DtbI_uHG66-cBk9grh5e0u6_LDt1k/edit)
- [ ] Forms and integrations 
	- [ ] Greenhouse (displaying current jobs) -- @bskinn 
	- [ ] Pipedrive (can use the form and the validation is working - no required fields can be left empty, email validation) -- @bskinn 
	- [ ] Mailchimp (submit form cannot be empty, email validation) -- @bskinn 
- [ ] Check hyperlinks (wonder if we should look into adding tests for this - through CI - though this might change the workflow with Vercel 🤔 )
	- [ ] Internal hyperlinks
	- [ ] External
- [ ] Check [loading speed](https://pagespeed.web.dev/)
- [ ] Components work as expected (not adding a comprehensive list here - carousels, links, buttons, dropdowns, etc.)

### Browsers

- [ ] Browser compatibility
	- [ ] Mobile
	- [ ] Chromium based web browsers
	- [ ] Safari
	- [ ] Firefox
	- [ ] Edge (?)

### CMS 

- [ ] Preview is working as expected
- [ ] Staging is content is updated once `Publish` in Storyblok is used
- [ ] Metadata for images, content types and others is adequately pre-defined (i.e. fields needed)

### Repository

- [ ] Technical documentation is included
	- [ ] Deployment and maintenance related
	- [ ] Dependencies and libraries are listed
	- [ ] How the repo is structured (where things live)
	- [ ] Coding standards are documented

	
### SEO

gabalafou:
@bskinn could I delegate to you the QA tasks under checking forms and integrations? I really need to focus my efforts on accessibility, URL routing, and some of the other QA tasks.

bskinn:
@gabalafou Sure, I can take those two.

- Greenhouse -- Can I just check this against the jobs listing on the current site?
- Pipedrive -- I'll also work with @bsodenkamp to make sure that the submitted data reaches the correct Pipedrive pipelines

[...]

PR #243, opened 2022-06-21
feat: Generate sitemaps using next-sitemap

gabalafou:
Add npm package `next-sitemap` along with configuration needed to generate site maps for Consulting and Labs sites. 

This is a first step to running pa11y against all of the pages on the websites. As such, it is a step toward closing the following issues: #152, #153, #154, and #156.

This code was largely patterned after code in the repo that generates the nx.dev website: https://github.com/nrwl/nx/blob/master/nx-dev/nx-dev/project.json#L19

## How to test locally

Consulting:

```sh
npx nx build consulting
npx nx sitemap consulting
```

Labs:

```sh
npx nx build labs
npx nx sitemap labs
```

## To do

- [ ] Make sitemap generation part of the build command so that when we run `nx build consulting`, sitemap.xml is output into the public folder
- [ ] Remove `<changefreq>` and `<priority>` from sitemap.xml if they have unintended consequences and are not needed
- [ ] Cherry on top: integrate the sitemap generation with the local dev server: `npm run start`


gabalafou:
Note to self: if this solution proves too inflexible or difficult to properly configure, it may be easier to create a sitemap component that works a bit like generateRSS.ts.

apps/consulting/project.json
```diff
@@ -15,6 +15,13 @@
         "production": {}
       }
     },
+    "sitemap": {
+      "executor": "@nrwl/workspace:run-commands",
+      "outputs": [],
+      "options": {
+        "command": "next-sitemap --config apps/consulting/next-sitemap.config.js"
```
gabalafou, line 22: @marcin-antczak-naturaily can we make this a part of the build step so that /sitemap.xml gets deployed with each Vercel deploy?

PR #248, opened 2022-06-22
[BLOG] add cxx numba interoperability

noatamir:
## Text styling

- [x] The blog is written with [plain language](https://www.plainlanguage.gov/guidelines/) (where relevant).
- [x] If there are headers, they use the proper header tags in order to do so (with only one level-one header).
- [x] All links describe where they link to (for example, check the [Quansight labs website](https://labs.quansight.org/)).
- [x] Any kind of styling that the author uses (for example, bold for emphasis) is consistent throughout the blog.

## Non-text contents

- [x] All content is represented as text (for example, images need alt text and videos need captions or descriptive transcripts).
- [x] If there are emojis, there are not more than three in a row.
- [x] Don't use [flashing gifs or videos](https://www.w3.org/TR/UNDERSTANDING-WCAG20/seizure-does-not-violate.html).
- [x] If it were to be read as plain text, the blog still makes sense and no information is missing.

## Links to other Quansight blogs:

no labs links found
no llc links found

PR #247, opened 2022-06-22
[BLOG] add portable-python-env

noatamir:
## Text styling

- [x] The blog is written with [plain language](https://www.plainlanguage.gov/guidelines/) (where relevant).
- [x] If there are headers, they use the proper header tags in order to do so (with only one level-one header).
- [x] All links describe where they link to (for example, check the [Quansight labs website](https://labs.quansight.org/)).
- [x] Any kind of styling that the author uses (for example, bold for emphasis) is consistent throughout the blog.

## Non-text contents

- [x] All content is represented as text (for example, images need alt text and videos need captions or descriptive transcripts).
- [x] If there are emojis, there are not more than three in a row.
- [x] Don't use [flashing gifs or videos](https://www.w3.org/TR/UNDERSTANDING-WCAG20/seizure-does-not-violate.html).
- [x] If it were to be read as plain text, the blog still makes sense and no information is missing.

## Links to other Quansight blogs:
no labs links found
no llc links found

PR #249, opened 2022-06-22
[BLOG] add the evolution of the scipy developer cli

noatamir:
## Text styling

## Non-text contents

no labs links found
no llc links found


noatamir:
🤔  it is not clear to me why the preview for this shows an empty blogs page (same for #247, #248).
Some guidance would be appreciated.

noatamir:
@gabalafou please check for A11y and merge CC @trallard

PR #260, opened 2022-06-28
[DESIGN] - Mobile Projects Page logo placement

kherma:


apps/labs/components/Projects/ProjectDescription.tsx
```diff
@@ -0,0 +1,33 @@
+import { FC } from 'react';
```
marcin-antczak-naturaily, line 1: No need to use FC here

apps/labs/components/Projects/ProjectHeadline.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC } from 'react';
```
marcin-antczak-naturaily, line 1: Also, no need to use FC (FC is useful only if component has {children})

marcin-antczak-naturaily (changes requested):

marcin-antczak-naturaily (approved):

PR #263, opened 2022-06-28
fix: handle variant of Storyblok token

marcin-antczak-naturaily:


gabalafou (commented):
Hey @marcin-antczak-naturaily, I left some questions on this pull request. As you can see, there's quite a bit I don't fully understand. I'm hoping that we will be able to clear these questions up in a call. :)

PR #265, opened 2022-06-28
fix: change name of page folder

marcin-antczak-naturaily:


kherma (approved):
Only remember to update Header / Footer links in SB to match new 'about-us' url

Issue #194, opened 2022-06-13
[DESIGN] - Change design of Consulting drop-down in the header


bskinn:
### What site is this for?

Quansight LLC

### Actual behavior

(1) 

The white background color and the black text color of the drop-down needs to be changed.

Desire:

* The green of the standard colors from the website theme for the background
* White for non-hovered text
* Black for hovered-over text

![image](https://user-images.githubusercontent.com/11325439/173932033-4df4745b-9a93-49d5-ab49-489aafadbbc9.png)

-----

(2)

The sub-items in the Consulting drop-down in the site header are center-justified within the drop-down, harming readability:

![image](https://user-images.githubusercontent.com/11325439/173410213-fbb8f7e0-2075-4ac1-ba54-6025eb1235ba.png)

Desire a left-align of the drop-down text, with a hanging indent on each service area name.

Adding a bold typeface to the "Consulting" item at the top of the header might also be good, to indicate that it is a top-level item, and the rest are sub-items.

bskinn:
An alternative solution might be to retain the center alignment, but to reduce the line spacing within each services area, to make the visual clustering of the lines more obvious.

gabalafou:
This issue has several sub-issues.

These sub-issues belong to Naturaily:

- Consulting menu items centered. This is non-standard. They should be left-aligned.
- Dropdown menu colors do not match design.

This sub-issue belongs to us:

- Special styling to sub-menu home page link. We could make it bold face. Another idea: the [accessible pattern that Deque uses](https://deque.com/) is to append the word "Overview" to the top item in each nav bar dropdown.

bskinn:
@katebrack @irinafumarel -- Thoughts on Gabriel's third sub-issue?

Katebrack:
Let's add the word "overview" to make it clear that's a separate page. On the current site, it's not clear that clicking "consulting" will take you to a separate page aside from what's in the dropdown menu. https://www.quansight.com/consulting @bskinn

trallard:
Just saw this - we decided to make the drop-down white since the combination green with white text is not accessible (does not meet the WCAG minimum contrast requirements)

bskinn:
FYI @irinafumarel.

Thanks for the heads up, @trallard!

As an alternative, would green with purple or black text meet the contrast spec?

bskinn:
Presuming this change moves forward, my preference is for black text on the green background.

Katebrack:
[...]

PR #274, opened 2022-06-30
fix: Adjust line spacing in Consulting drop down in header

bskinn:
Increase `gap` class property for line spacing between list elements.

Decrease `leading` class property for line spacing within each list element.

A further change after #194.


bskinn:
This is a small enough change that I'm comfortable merging it without review, during this hustle up to launch.

PR #288, opened 2022-07-01
fix: First batch of accessibility fixes for Consulting homepage

gabalafou:
This is my first batch of tweaks to improve the accessibility of the Consulting homepage. Note that many of these fixes will carry over to other Consulting pages, as well as Labs pages, but they were all derived from manually inspecting the Consulting homepage.

There are a few issues on the homepage that I'm not tackling in this PR:

- There are three links that currently just say "Learn more." Fixing these either requires a copy change or a non-trivial change in Storyblok and/or the codebase.
- Not sure the following should be an h3: "Turning data into insights with open source software."
- The skip link does not scroll past the top nav bar. 
- I think clicking "Consulting" in the nav bar should make the menu stay open until clicking elsewhere (don't close menu on mouse out).

### DIFF
--- a/apps/consulting/components/Board/BoardItem.tsx
+++ b/apps/consulting/components/Board/BoardItem.tsx
@@ -30,23 +30,23 @@ const BoardItem: FC<TBoardItemProps & { classNameBorder: string }> = ({
           priority
         />
       </div>
-      <h3 className="text-[2.2rem] font-extrabold leading-[3rem] sm:mt-[2.8rem] sm:mb-[2.2rem] xl:mt-[3.6rem] xl:mb-[2.2rem] font-heading">
-        {title}
-      </h3>
-      <div className="flex">
-        <Link href={linkUrl}>
-          <a className="flex gap-3 justify-center items-center mx-auto w-auto text-[1.6rem] font-bold leading-[3.7rem]">
+      <Link href={linkUrl}>
+        <a>
+          <div className="text-[2.2rem] font-extrabold leading-[3rem] sm:mt-[2.8rem] sm:mb-[2.2rem] xl:mt-[3.6rem] xl:mb-[2.2rem] font-heading">
+            {title}
+          </div>
+          <div className="flex gap-3 justify-center items-center mx-auto w-auto text-[1.6rem] font-bold leading-[3.7rem]">
             {linkTitle}
             <Picture
               imageSrc="/board/board-btn-arrow.svg"
-              imageAlt="arrow icon"
+              imageAlt=""
               width={12}
               height={17}
               priority
             />
-          </a>
-        </Link>
-      </div>
+          </div>
+        </a>
+      </Link>
     </div>
   );
 };
--- a/apps/consulting/components/Board/BoardItem.tsx
+++ b/apps/consulting/components/Board/BoardItem.tsx
@@ -12,7 +12,6 @@ const BoardItem: FC<TBoardItemProps & { classNameBorder: string }> = ({
   linkTitle,
   linkUrl,
   imageSrc,
-  imageAlt,
   classNameBorder,
 }) => {
   return (
@@ -22,16 +21,17 @@ const BoardItem: FC<TBoardItemProps & { classNameBorder: string }> = ({
         classNameBorder,
       )}
     >
-      <div className="relative h-32">
-        <Picture
-          imageSrc={imageSrc}
-          imageAlt={imageAlt}
-          layout="fill"
-          priority
-        />
-      </div>
       <Link href={linkUrl}>
         <a>
+          <div className="relative h-32">
+            <Picture
+              imageSrc={imageSrc}
+              // Alt text deliberately empty because the images used here are decorative.
+              imageAlt=""
+              layout="fill"
+              priority
+            />
+          </div>
           <div className="text-[2.2rem] font-extrabold leading-[3rem] sm:mt-[2.8rem] sm:mb-[2.2rem] xl:mt-[3.6rem] xl:mb-[2.2rem] font-heading">
             {title}
           </div>
--- a/apps/consulting/components/Board/decorations/HeaderDecoration.tsx
+++ b/apps/consulting/components/Board/decorations/HeaderDecoration.tsx
@@ -6,7 +6,7 @@ export const HeaderDecoration: FC = () => (
   <div className="hidden absolute top-[6%] left-[6%] w-[16.6rem] h-[12.5rem] lg:block xl:top-[15%] xl:left-[2%]">
     <Picture
       imageSrc="/board/board-header-icon.svg"
-      imageAlt="header geometric decoration"
+      imageAlt=""
       layout="responsive"
       width={166}
       height={125}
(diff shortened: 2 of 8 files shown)


libs/shared/ui-components/src/Footer/FooterColumnProvider/FooterNavigation.tsx
```diff
@@ -12,14 +12,18 @@ export const FooterNavigation: FC<TFooterNavigationProps> = ({
...
-    <nav className="flex flex-col">
-      {links.map(({ linkText, linkUrl, _uid }) => (
-        <Link href={`/${linkUrl.cached_url}`} key={_uid}>
-          <a className="text-[1.4rem] font-normal leading-[3.3rem]">
-            {linkText}
-          </a>
-        </Link>
-      ))}
-    </nav>
+    <div className="flex flex-col">
```
gabalafou, line 15: Per guidance from WebAIM, I decided that [none of the links in the footer should be enclosed in a `nav` tag](https://webaim.org/techniques/semanticstructure/).

I wasn't sure whether they should be wrapped in a list (still not sure), but I just copied a11yproject.com on this one.

libs/shared/ui-components/src/Header/HeaderDesktop/HeaderDesktopNavigationProvider/HeaderDesktopDropdown.tsx
```diff
@@ -75,7 +75,7 @@ export const HeaderDesktopDropdown: FC<THeaderDesktopDropdownProps> = ({
         <div className="absolute top-full right-0 z-20 pt-[2.4rem] w-max max-w-[25rem]">
           <ul
             id="options"
-            className="flex flex-col gap-[1.8rem] justify-between items-start p-[2.8rem] min-w-[19.8rem] text-left text-black bg-white/90"
+            className="flex flex-col gap-[1.8rem] justify-between items-start p-[2.8rem] min-w-[19.8rem] text-left text-black bg-white"
```
gabalafou, line 78: I'm not 100% sure but I strongly suspected that the 90% opacity in the dropdown was creating contrast issues, or strong potential for contrast issues.

libs/shared/ui-components/src/Header/HeaderDesktop/HeaderDesktopNavigationProvider/HeaderDesktopDropdownLink.tsx
```diff
@@ -19,7 +19,7 @@ export const HeaderDesktopDropdownLink: FC<THeaderDesktopDropdownLinkProps> = ({
 
   return (
     <button
-      className="text-[1.7rem] font-normal leading-[2.225rem] text-left capitalize transition-colors motion-reduce:transition-none ease-in-out hover:text-green font-heading focus:text-green"
+      className="text-[1.7rem] font-normal leading-[2.225rem] text-left capitalize transition-colors motion-reduce:transition-none ease-in-out hover:text-darkgreen focus:text-darkgreen font-heading"
```
gabalafou, line 22: Here I changed the text color from green to a color I aliased to "darkgreen" (558300) to make it AA compliant.

gabalafou (commented):
done self-reviewing

apps/consulting/components/Board/BoardItem.tsx
```diff
@@ -30,23 +30,23 @@ const BoardItem: FC<TBoardItemProps & { classNameBorder: string }> = ({
...
       </div>
-      <h3 className="text-[2.2rem] font-extrabold leading-[3rem] sm:mt-[2.8rem] sm:mb-[2.2rem] xl:mt-[3.6rem] xl:mb-[2.2rem] font-heading">
-        {title}
-      </h3>
-      <div className="flex">
-        <Link href={linkUrl}>
-          <a className="flex gap-3 justify-center items-center mx-auto w-auto text-[1.6rem] font-bold leading-[3.7rem]">
+      <Link href={linkUrl}>
+        <a>
+          <div className="text-[2.2rem] font-extrabold leading-[3rem] sm:mt-[2.8rem] sm:mb-[2.2rem] xl:mt-[3.6rem] xl:mb-[2.2rem] font-heading">
```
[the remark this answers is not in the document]
isabela-pf: @gabalafou tested this with a screen reader, and as long as the image remains decorative the experience for this is far more descriptive than it was when the links were only "Learn More" several times in a row. All appears to be working wonderfully!

Another plus I find with this approach is that the clickable region is now larger which can make it clearer this is a link and not just a table.

apps/consulting/components/Board/BoardItem.tsx
```diff
@@ -30,23 +30,23 @@ const BoardItem: FC<TBoardItemProps & { classNameBorder: string }> = ({
...
+            {title}
+          </div>
+          <div className="flex gap-3 justify-center items-center mx-auto w-auto text-[1.6rem] font-bold leading-[3.7rem]">
             {linkTitle}
             <Picture
               imageSrc="/board/board-btn-arrow.svg"
-              imageAlt="arrow icon"
+              imageAlt=""
```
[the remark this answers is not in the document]
isabela-pf: I agree! This follows the same logic as [my previous reply](https://github.com/Quansight/Quansight-website/pull/288#discussion_r913140980).

apps/consulting/components/Board/decorations/HeaderDecoration.tsx
```diff
@@ -6,7 +6,7 @@ export const HeaderDecoration: FC = () => (
   <div className="hidden absolute top-[6%] left-[6%] w-[16.6rem] h-[12.5rem] lg:block xl:top-[15%] xl:left-[2%]">
     <Picture
       imageSrc="/board/board-header-icon.svg"
-      imageAlt="header geometric decoration"
+      imageAlt=""
```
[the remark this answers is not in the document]
isabela-pf: I agree! This follows the same logic as [my previous reply](https://github.com/Quansight/Quansight-website/pull/288#discussion_r913140980).

libs/shared/ui-components/src/Header/HeaderDesktop/HeaderDesktopNavigationProvider/HeaderDesktopDropdown.tsx
isabela-pf: Looking at it, because the text and menu background are high contrast they do still pass contrast thresholds at a 90% opacity. However, I still don't consider it the most sustainable accessible approach since the changing background below the menu could interfere with it in the future. I prefer not to recommend approaches that put more work on people in the future where possible.

From a design perspective, I'd also advocate for the 100% opacity menu in this case. I can't find less than 100% opacity used almost at all in this website design, so I find full opacity more aligned with the system anyways.

libs/shared/ui-components/src/Header/HeaderDesktop/HeaderDesktopNavigationProvider/HeaderDesktopDropdownLink.tsx
isabela-pf: I'm pretty sure that the green used for this hover state is a brand color, so we probably need to talk to the design team or come up with an alternative approach that still follows the brand guidelines. Maybe there's another green (or other color) in the brand that works better? Though since we're using it to work on both very dark and very light backgrounds in this case, I am not confident the same green will have enough contrast for both.

I have a few other ideas for long term! I know with the timeline they are unlikely to happen in this PR.
- Since there's only one menu that has that drop down, why not come up with another way to represent that information. Maybe the landing page for the section has all those options listed at the top?
- Don't rely on color (or solely on color) to signal the hover state. Use an icon or an underline or something similar.
- Change the background color of the menu to be dark, like the main navigation elsewhere.
- Work with the brand stewards to come up with an extension of the existing color palette that gives a greater range of values to work with in this situation. Maybe there is a dark equivalent of the brand green they'd like to incorporate.
- A combination of the above options.
- Experiment with CSS text shadow options. I do not recommend this approach when this website does not have other soft or low opacity treatments, but it is an option to try if nothing else works.

PR #292, opened 2022-07-04
fix: staff spotlight names

kherma:


libs/shared/ui-components/src/Team/Team.tsx
```diff
@@ -17,17 +19,34 @@ export const Team: FC<TTeamProps> = ({
...
+    (prevValue, { displayName, firstName, lastName }) => {
+      const rawName =
+        displayName === TeamDisplay.Full
+          ? `${firstName} ${lastName}`
+          : firstName;
+
+      return rawName?.length <= 15 && prevValue === false ? false : true;
+    },
+    false,
+  );
```
mstoczkowski, line 32: Generally it is ok but i would recommend to move it out to separate function

mstoczkowski (approved):

PR #297, opened 2022-07-04
fix: Fix book a call link in mobile nav menu

bskinn:
Needed to be switched from `/about` to `/about-us`.

No associated issue.


bskinn:
Fix is confirmed by manual testing of menu in mobile view. Merging without review.

PR #298, opened 2022-07-04
[BUG] - Board List on Consulting page renders wider than browser window width

kherma:


bskinn:
I think this is OK to merge to `develop` too, @kherma -- we definitely want this in `main` ASAP.

mstoczkowski (approved):

PR #315, opened 2022-07-05
fix: differentiate Learn More links on Consulting home

gabalafou:
This PR is intended to only affect the Consulting home page (I think that's the only place where the StickyNote component is used with a Learn More link). 

Instead of having two links that appear in the links list as:

Learn More 
Learn More

They will appear as:

Consulting Learn More
Labs Learn More

Important: this does not affect the __visual__ representation of links, only the way they are represented to assistive tech (screen readers).


bskinn:
@gabalafou Does this change how the buttons render in a standard browser view? Or just for e.g. screen readers?

gabalafou:
Should be just screen readers

gabalafou:
The visible text for these links has already changed on the Consulting homepage. They now read "See Consulting Areas" and "Visit Quansight Labs", so this pull request is no longer needed.

PR #252, opened 2022-06-23
feat: set up URL rewrites for Consulting and Labs

gabalafou:
Note: the desired behavior is described better by [#342](https://github.com/Quansight/Quansight-website/pull/324), but as I have detailed in inline comments on that PR, it doesn't work, so this is what we have to live with for now.

Consulting:

- Any URL /post/title-of-blog-post that does not match a blog post on the new site will get a 404.
- Any other URL that's not handled by the new site will result in a 404.

Labs: 

- Any URL like /blog/2022/01/title-of-blog-post will go to the old site.
- Any other URL that's not handled by the new site should fallback to the old site.

## Testing

URLs to test for Consulting:

| Description | Behavior | URL |
|--------|-------|-------|
| Migrated blog post | Render post in new site |  https://quansight-consulting-staging-git-fallback-urls-quansight.vercel.app/post/up-and-running-with-prefect | 
| Unmigrated blog post | 404 on new site | https://quansight-consulting-staging-git-fallback-urls-quansight.vercel.app/post/scipyla-scientific-python-in-latin-america |

URLs to test for Labs:

| Description | Behavior | URL |
|--------|-------|-------|
| Old year/month URL of migrated blog post | Render post in old site | https://quansight-labs-staging-git-fallback-urls-quansight.vercel.app/blog/2020/06/portable-python-env |
| Old year/month URL of unmigrated blog post | Render post in old site | https://quansight-labs-staging-git-fallback-urls-quansight.vercel.app/blog/2019/05/labs-update-april-highlights | 
| New URL of migrated blog post | Render post in new site | https://quansight-labs-staging-git-fallback-urls-quansight.vercel.app/blog/portable-python-env 
| New URL of unmigrated blog post | 404 from old site | https://quansight-labs-staging-git-fallback-urls-quansight.vercel.app/blog/labs-update-april-highlights |

Here are some old URLs to also test that the fallback works. These should render pages from the old site (while staying at the same URL):

- https://quansight-labs-staging-git-fallback-urls-quansight.vercel.app/2022
- https://quansight-labs-staging-git-fallback-urls-quansight.vercel.app/archive
- https://quansight-labs-staging-git-fallback-urls-quansight.vercel.app/archive.html
- https://quansight-labs-staging-git-fallback-urls-quansight.vercel.app/search?q=python

PR #324, opened 2022-07-07
Why doesnt this work

gabalafou:
I created this draft pull request just to show some code that I wrote that didn't work. I have described with in-line comments what did not work.

The code in this PR is what I wanted to run on the websites, but I had to make modifications because this code didn't work. You can see the compromises I made in #252.

## Desired URL behavior

Consulting:

- Any URL /post/title-of-blog-post that does not match a blog post on the new site, will get a temporary 307 redirect to /page-unavailable
- Any other URL that's not handled by the new site will result in a 404.

Labs: 

- Any URL like /blog/2022/01/title-of-blog-post will get redirected to /blog/title-of-blog-post. If that URL does not match a blog post currently in the site, it will fallback to the old Labs site.
- Any other URL that's not handled by the new site should fallback to the old site.

URLs to test for Consulting:

| Description | Behavior | URL path |
|--------|-------|-------|
| Migrated blog post | Render post in new site |  /post/up-and-running-with-prefect | 
| Unmigrated blog post | 404 on new site | /post/scipyla-scientific-python-in-latin-america |

**Note**: All of these URL paths should be tested against new site. In other words they should hit the Next.js app on Vercel servers. For example: https://quansight-consulting.vercel.app/post/up-and-running-with-prefect.

URLs to test for Labs:

| Description | Behavior | URL path |
|--------|-------|-------|
| Old year/month URL of migrated blog post | Render post in new site | /blog/2020/06/portable-python-env |
| Old year/month URL of unmigrated blog post | Render post in old site | /blog/2019/05/labs-update-april-highlights | 
| New URL of migrated blog post | Render post in new site | /blog/portable-python-env 
| New URL of unmigrated blog post | 404 from old site | /blog/labs-update-april-highlights |

Here are some old URLs to also test that the fallback works. These should render pages from the old site (while staying at the same URL):

- /2022
- /archive
- /archive.html
- /search?q=python

**Note**: All of these URL paths should be tested against new site. In other words they should hit the Next.js app on Vercel servers. For example: https://quansight-labs.vercel.app/blog/2020/06/portable-python-env


apps/consulting/next.config.js
```diff
@@ -47,16 +47,30 @@ const nextConfig = {
...
+          // https://plausible.io/docs/proxy/guides/vercel
+          source: '/plausible/js/script.js',
+          destination: 'https://plausible.io/js/script.js',
+        },
+        {
+          source: '/plausible/api/event',
+          destination: 'https://plausible.io/api/event',
+        },
+      ],
+      fallback: [
```
gabalafou, line 67: What I expect this to do:

- User goes to http://quansight-consulting.vercel.app/post/this-post-has-not-yet-been-migrated.
- Url in the browser stays the same.
- User sees the `page-unavailable` page currently in the LLC Storyblok (simple page that says we're working on migrating the content).

What actually happens:

- All the same things as above.
- Page attempts to fetch a non-existent JSON file: `this-post-has-not-yet-been-migrated.json`.
- Because there's no JSON at the end point, the code that fetched the JSON throws an error.
- Some Next.js or React.js machinery catches the error and renders an error message on the page.

apps/labs/next.config.js
```diff
@@ -30,18 +30,78 @@ const nextConfig = {
...
+        {
+          source: '/plausible/api/event',
+          destination: 'https://plausible.io/api/event',
+        },
+      ],
+      fallback: [
+        // These rewrites are checked after both pages/public files and dynamic
+        // routes are checked.
+        {
+          source: '/blog/:slug',
```
gabalafou, line 70: For some reason this works locally (localhost:4200) but not on vercel.app.

Here's what happens locally:

- I visit the URL to a page that hasn't yet been migrated, such as `blog/2019/05/labs-update-april-highlights`.
- Local server follows redirect rule above and redirects the user's browser to `/blog/labs-update-april-highlights?yyyy=2019&mm=05.
- There is no blog post currently at that URL on localhost, so local server then proxies the request as expected, and renders the `labs-update-april-highlights` blog post from the old site, in other words it shows the exact same page you would get if you get the page from Netlify at this URL: https://quansight-labs.netlify.app/2019/05/blog/labs-update-april-highlights/. The URL in the browser, of course, stays the same.

What happens in production:

- I visit the URL, for example: https://quansight-labs-staging-git-why-doesnt-this-work-quansight.vercel.app/blog/2019/05/labs-update-april-highlights
- It gets redirected as expected.
- But, instead of proxying the old blog post from Netlify, we get the old Netlify 404.

I don't know why this works locally but fails on Vercel.app.

apps/consulting/next.config.js
gabalafou: What I don't understand is why Next.js attempts to fetch that JSON file. The documentation for `getStaticPaths` seems to indicate that if you have `fallback: false`—like we do for all post/[slug] dynamic route—then data-fetching shouldn't happen.

gabalafou: I wondered if maybe this was something built in to the way that Next.js works or not. So I created a small test site in which I mimicked several things from this Next.js app:

- [Same Next.js version ](https://github.com/gabalafou/nextjs/blob/main/package.json#L10)
- Same dynamic route. Compare Consulting's [pages/post/[slug].tsx](https://github.com/Quansight/Quansight-website/blob/develop/apps/consulting/pages/post/%5Bslug%5D.tsx) to my test site's [pages/[slug].jsx](https://github.com/gabalafou/nextjs/blob/main/pages/%5Bslug%5D.jsx)
- [Same fallback rewrite](https://github.com/gabalafou/nextjs/blob/main/next.config.js#L6)

So what happens? On my test site, it does exactly what I expected the rewrite to do on the Consulting site. If you visit a URL on my test site, such as [/does-not-exist](https://nextjs-pi-one-76.vercel.app/does-not-exist) that does not match one of the paths return by `getStaticPaths`, then the server falls back to page returned from /page-unavailable. The client app does not attempt to fetch a non-existing JSON file that 404s, and it does not throw an error.

bskinn: It appears that the JSON error is coming about because the 404 page is being (i) erroneously served, _at all_, and (ii) erroneously served _as a JSON payload_.

From the browser console:

```
SyntaxError: Unexpected token < in JSON at position 0
```

From the JSON payload:

```
<!DOCTYPE html><html lang="en" class="text-[62.5%]"><head><meta charSet="utf-8"/><meta name="viewport" content="width=device-width, initial-scale=1"/><title>404: This page could not be found</title> ...
```

----

It also appears to be specific to the `/post/:slug` fallback rewrite, as a [similar rewrite](https://github.com/Quansight/Quansight-website/commit/83d91404d465ee149e071ca8117fec31c271d66e) for a plain, static page [works correctly](https://quansight-consulting-published-602d508u5-quansight.vercel.app/test-missing).

PR #337, opened 2022-07-08
fix: update Plausible code snippets

gabalafou:
- Replace placeholder domains with our actual domains
- Change subdirectory name from plausible to p7e
- Add /opt-out-analytics pages to Consulting and Labs sites

### DIFF
--- a/apps/consulting/next.config.js
+++ b/apps/consulting/next.config.js
@@ -56,11 +56,16 @@ const nextConfig = {
         {
           // Proxying Plausible through Vercel:
           // https://plausible.io/docs/proxy/guides/vercel
-          source: '/plausible/js/script.js',
+
+          // `p7e` stands for Plausible. Why not just use `plausible`? The docs
+          // say: Choose a generic or irrelevant name for the subdirectory. If
+          // you choose something like analytics or plausible, it might get
+          // blocked in the future.
+          source: '/p7e/js/script.js',
           destination: 'https://plausible.io/js/script.js',
         },
         {
-          source: '/plausible/api/event',
+          source: '/p7e/api/event',
           destination: 'https://plausible.io/api/event',
         },
       ],
--- a/apps/consulting/pages/_app.tsx
+++ b/apps/consulting/pages/_app.tsx
@@ -1,13 +1,26 @@
 import { AppProps } from 'next/app';
+import Script from 'next/script';
 
 import { Meta } from '@quansight/shared/ui-components';
+
 import './styles.css';
 
 function CustomApp({ Component, pageProps }: AppProps): JSX.Element {
+  const consultingDomain = 'quansight.com';
   return (
     <>
       <Meta />
       <Component {...pageProps} />
+
+      {typeof window !== 'undefined' &&
+        window.location.hostname === consultingDomain && (
+          // For more info about this script, see note in consulting/next.config.js
+          <Script
+            data-domain={consultingDomain}
+            data-api="/p7e/api/event"
+            src="/p7e/js/script.js"
+          />
+        )}
     </>
   );
 }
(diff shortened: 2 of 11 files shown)


apps/consulting/next.config.js
```diff
@@ -56,11 +56,16 @@ const nextConfig = {
         {
           // Proxying Plausible through Vercel:
           // https://plausible.io/docs/proxy/guides/vercel
-          source: '/plausible/js/script.js',
+
+          // `p7e` stands for Plausible. Why not just use `plausible`? The docs
+          // say: Choose a generic or irrelevant name for the subdirectory. If
+          // you choose something like analytics or plausible, it might get
+          // blocked in the future.
+          source: '/p7e/js/script.js',
```
gabalafou, line 64: There's a good discussion about [ad blockers and proxying](https://plausible.io/docs/proxy/introduction) in the Plausible docs. 

I think I mostly agree with the Plausible point of view, although I think there is a risk here of disrespecting users who are not blocking for other reasons besides privacy. For example, perhaps they are attempting to limit all extraneous requests because they are working in a constrained environment, perhaps with limited bandwidth or power.

gabalafou (commented):
Done self-reviewing.

apps/labs/pages/_document.tsx
```diff
@@ -14,11 +14,12 @@ const Document: FC = () => {
...
         <NextScript />
+        {/* For more info about this script, see note in labs/next.config.js */}
         <script
           defer
-          data-domain="yourdomain.com" // TODO to be changed before deploy
-          data-api="/plausible/api/event"
-          src="/plausible/js/script.js"
+          data-domain="labs.quansight.org"
+          data-api="/p7e/api/event"
+          src="/p7e/js/script.js"
```
trallard, line 22: Since I have not looked at the docs in detail I am assuming this is the correct end-point to the API

apps/consulting/next.config.js
trallard: Ah this answers my question about the API - 👍🏽

apps/consulting/pages/_document.tsx
```diff
@@ -14,11 +14,12 @@ const Document: FC = () => {
       <body>
         <Main />
         <NextScript />
+        {/* For more info about this script, see note in consulting/next.config.js */}
         <script
           defer
-          data-domain="yourdomain.com" // TODO to be changed before deploy
-          data-api="/plausible/api/event"
-          src="/plausible/js/script.js"
+          data-domain="quansight.com"
```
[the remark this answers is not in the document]
trallard: From reading through this - I do not think this will not track Vercel I suppose we will see 🤷🏽‍♀️

trallard (approved):

apps/consulting/pages/_document.tsx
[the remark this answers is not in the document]
gabalafou: Okay, I just checked this. I loaded the Vercel preview URL, which is on vercel.app domain, for this PR, and the Plausible dashboard went from showing 0 visitors to labs.quansight.org to showing 1 visitor:



So I need to update this PR to only load Plausible when the user it at labs.quansight.org. 

Something to look into in the future would be if we want to have a staging site in Plausible so we can test our analytics from staging before pushing changes to production.

apps/consulting/pages/_app.tsx
```diff
@@ -1,13 +1,26 @@
...
 
 function CustomApp({ Component, pageProps }: AppProps): JSX.Element {
+  const consultingDomain = 'quansight.com';
   return (
     <>
       <Meta />
       <Component {...pageProps} />
+
+      {typeof window !== 'undefined' &&
+        window.location.hostname === consultingDomain && (
```
gabalafou, line 16: The idea here is that we don't want to send events to Plausible analytics when we're testing things on the vercel.app domains. Plausible excludes localhost by default but doesn't know how to exclude other domains.

PR #338, opened 2022-07-08
fix: remove link to book a call from Labs mobile menu

gabalafou:
### How to test this PR

Consulting 

- Go to -live preview URL on mobile browser (or emulate mobile in desktop browser).
- Open hamburger menu.
- Verify that the "book a call" link is in the menu and that clicking it takes you to the contact form.

Labs:

- Go to -live preview URL on mobile browser.
- Open hamburger menu.
- Verify that the "book a call" link is no longer in the menu.

### DIFF
--- a/libs/shared/ui-components/src/Header/HeaderMobile/HeaderMobileNavigation.tsx
+++ b/libs/shared/ui-components/src/Header/HeaderMobile/HeaderMobileNavigation.tsx
@@ -2,6 +2,8 @@ import { FC } from 'react';
 
 import clsx from 'clsx';
 
+import { DomainVariant } from '@quansight/shared/types';
+
 import { HeaderMobileBookingLink } from './HeaderMobileBookingLink';
 import { HeaderMobileNavigationProvider } from './HeaderMobileNavigationProvider/HeaderMobileNavigationProvider';
 import { THeaderMobileNavigationProps } from './types';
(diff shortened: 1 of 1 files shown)


gabalafou:
@kherma, in theory once this PR is merged, we can remove `bookACallLinkText` field from the Labs header schema, correct?

Another question: editing the schema in under the Labs Storyblok space does not change the schema under Consulting Storkblok space, does it? These are completely separate, correct?

Don't worry I'm not planning on editing the schema much or at all. I'm asking these questions to help me better understand the system.

kherma:
Okay, so theoretically that will fix it. But you should not delete the `bookACallLinkText` from labs schema after merge.

## Graphql Queries

To fetch the data from SB API, you need to create query schema. Now, as we use typescript we want to add types of date we want to fetch. But we have 3 types of components: Labs only, LLC only and shared. And because the data that comes for SB is huge, we use codegen to generate types for us. Queries for all shared components are stored in `libs/shared/storyblok-sdk/api/queries` and in `apps/*/api/queries` we store queries specific to LLC / Labs. Query for Header is shared between LLC & Labs as they were the same on the design.

## Modifying Queries

If you want to modify db schema to differentiate LLC component from Labs you can create new query in e.g. apps/labs... and run codegen for that domain. Because if you only delete the field from SB schema, you still trying to fetch that field with your query:

broken-query
graphql-error

kherma (approved):

PR #342, opened 2022-07-12
feat: Put project title in heading id for anchor linkability

gabalafou:
This PR will allow linking directly to a project on the project page. For example, https://labs.quansight.org/projects#Zarr.

Improvements for now or the future:

- Canonicalize? Replace spaces with dashes. Make lowercase? 
- Expand project section when user jumps to it via anchor link.
- Add clickable anchor-link next to heading. Right now there's no way to discover these heading links without looking at the source code and knowing that you can hash-link directly to headings with ids.


trallard:
Verified and this does work for https://labs-git-project-headings-ids-quansight.vercel.app/projects#Zarr

> Improvements for now or the future:

> Canonicalize? Replace spaces with dashes. Make lowercase?

Definitely - we should aim to slugify everything (dashes-no-spaces-and-lowercase)

The other two suggestions re expanding and adding the anchor-link also seem like good improvements 👍🏽

trallard (approved):

PR #354, opened 2022-07-15
fix: Switch getPageItems Apollo query fetch policy to network only

### LINKED ISSUE #353: [BUG] - Non-deterministic `TypeError: Cannot read property 'firstName' of undefined` during LLC build
### What site is this for?

Quansight LLC

### Expected behavior

_No response_

### Actual behavior

On some builds of the LLC website, the build fails with `TypeError: Cannot read property 'firstName' of undefined` errors for some, but not all, published blog posts. Here are three example failing builds: [one](https://vercel.com/quansight/quansight-consulting-published/52w6VeoNosde1W4jhFegtCdWbWg6), [two](https://vercel.com/quansight/quansight-consulting-published/48TY1oL781rmWE5LKgYVdA6Kp82D), [three](https://vercel.com/quansight/quansight-consulting-published/52w6VeoNosde1W4jhFegtCdWbWg6).

The error/failure is not deterministic. Redeploying a build that fails from this error can sometimes succeed in the future, though it will often take 5+ redeployments before a successful build.

For reference, a full traceback for a representative error:

```
Error occurred prerendering page "/post/extending-numba-types-for-clean-fast-code". Read more: https://nextjs.org/docs/messages/prerender-error
--
14:10:42.587 | TypeError: Cannot read property 'firstName' of undefined
14:10:42.587 | at /vercel/path0/dist/apps/consulting/.next/server/chunks/629.js:245:141
14:10:42.587 | at Array.map (<anonymous>)
14:10:42.587 | at getBlogArticlesProps (/vercel/path0/dist/apps/consulting/.next/server/chunks/629.js:232:65)
14:10:42.587 | at getLibraryTiles (/vercel/path0/dist/apps/consulting/.next/server/chunks/629.js:299:83)
14:10:42.587 | at getStaticProps (/vercel/path0/dist/apps/consulting/.next/server/pages/post/[slug].js:174:123)
14:10:42.587 | at processTicksAndRejections (internal/process/task_queues.js:95:5)
14:10:42.587 | at async renderToHTML (/vercel/path0/node_modules/next/dist/server/render.js:492:20)
14:10:42.587 | at async /vercel/path0/node_modules/next/dist/export/worker.js:253:36
14:10:42.587 | at async Span.traceAsyncFn (/vercel/path0/node_modules/next/dist/trace/trace.js:79:20)


```

----

Originally, I thought there was something wrong with a specific team member in `/team`. In the leadup to launch, when I was in the course of trying to publish all of the migrated blog posts, the build error would occur whenever I had either of the two posts by Adam Lewis, "Panel/Holoviews Learning Aid" and/or "Spatial Filtering at Scale with Dask and Spatialpandas", set in the *Published* state. If both were *Unpublished*, the build would consistently succeed.

To emphasize again: the `TypeError` would occur on *more posts* than just these two posts I thought were problematic.

This assumption, that Adam Lewis's author entry was the problem, was reinforced when I switched the author of one of these posts to Dharhas, set that post to *Published*, and observed successful builds. If I then switched back to Adam Lewis as author, the build failure would recur.

----

Later on, after @kherma's fix for #310 was implemented, the same build errors started to occasionally occur again, even with Adam Lewis's posts set as *Unpublished*. I noted this in https://github.com/Quansight/Quansight-website/pull/310#issuecomment-1174962115 and the following conversation.  It happened infrequently enough, though, that I decided it wasn't worth trying to fix before launch -- I would just manually kick off redeploys as needed until I got a successful build.

----

One final observation: as best I know, these build errors **ONLY** occur on builds that are configured to use **only *Published* Storyblok content** -- `-staging` builds triggered on the exact same Github code and Storyblok content, which pick up **both *Published* and *Unpublished* content**, do not experience this error. As an example, see the following two builds, from one point in the midstream of @gabalafou's diagnosis efforts for the problem (commit 78e192e):

- [`quansight-consulting-published` build](https://vercel.com/quansight/quansight-consulting-published/52w6VeoNosde1W4jhFegtCdWbWg6) -- only *Published* content, error occurs
- [`quansight-consulting-staging` build](https://vercel.com/quansight/quansight-consulting-staging/D3cYBqjMc2GanXzufKQkR6ED9PAF) -- both *Published* and *Unpublished* content, **NO** error occurs

----

The non-determinism here makes me think that this may be a race condition in the pre-rendering step: somehow, author metadata information is being requested from some object(s) that *will* hold it in the future, but that have not yet been populated at the time of these errors. Given the appearance of `getLibraryTiles` in the traceback, it seems likely that the problem is coming from the library-population code for `/library`, where the page collects the information it needs in order to show the grid of library items.

I'm at a loss for why the error would only occur for builds with only *Published* content. Perhaps there's something in the logic where the Next.js code is querying the `/post` items for *(Un)published* status that is contributing to a possible race condition? Or triggering early/late population of the author metadata?

If it _is_ a race condition, perhaps putting some sort of retryer or `setTimeout()` on the [call to `getAuthorName()` in `getBlogArticlesProps.ts`](https://github.com/Quansight/Quansight-website/blob/1abd992ecd6800b0fb8dc04169c97b5c0dd5702c/apps/consulting/utils/getLibraryTiles/getBlogArticlesProps.ts#L22-L25), and possibly also to the similar call in [`getLibraryLinksProps.ts`](https://github.com/Quansight/Quansight-website/blob/1abd992ecd6800b0fb8dc04169c97b5c0dd5702c/apps/consulting/utils/getLibraryTiles/getLibraryLinksProps.ts#L14-L17), might help? These would probably only be band-aids, though, not solutions.

_No response_

### Anything else?

_No response_

bskinn:
If it works, closes #353.

~Adds explicit handling of case where `blogArticleBodyItem.author` is a Promise, instead of an object with the needed author data.~

Change the [Apollo query fetch policy](https://www.apollographql.com/docs/react/data/queries/#supported-fetch-policies) from `cache-first` to `network-only` for `getPageItems`. It seems that the author data in the cache could sometimes be stale -- or perhaps cross-contaminated by concurrent builds? -- leading to garbage data (in the form of a UID) being returned/stored.

### DIFF
--- a/apps/consulting/utils/getLibraryTiles/getBlogArticlesProps.ts
+++ b/apps/consulting/utils/getLibraryTiles/getBlogArticlesProps.ts
@@ -12,6 +12,10 @@ export const getBlogArticlesProps = (blogArticles: PageItems): TTiles =>
       ),
     };
 
+    if ( typeof blogArticleBodyItem.author.content === 'undefined' ) {
+      console.log(blogArticleBodyItem);
+    }
+
     /* blogArticleBodyItem.author could be a Promise */
     let author = '';
     let author_obj = blogArticleBodyItem.author;
--- a/apps/consulting/utils/getLibraryTiles/getBlogArticlesProps.ts
+++ b/apps/consulting/utils/getLibraryTiles/getBlogArticlesProps.ts
@@ -13,16 +13,19 @@ export const getBlogArticlesProps = (blogArticles: PageItems): TTiles =>
     };
 
     if ( typeof blogArticleBodyItem.author.content === 'undefined' ) {
-      console.log(blogArticleBodyItem);
+      console.log(`Title: ${blogArticleBodyItem.postTitle}`);
+      console.log(`Incoming Author: ${blogArticleBodyItem.author}`);
     }
 
-    /* blogArticleBodyItem.author could be a Promise */
+    // blogArticleBodyItem.author could be a Promise
     let author = '';
     let author_obj = blogArticleBodyItem.author;
     if (typeof author_obj === 'object' && typeof author_obj.then === 'function') {
       let promise_author = {author: ''};
-      author_obj.then( auth => promise_author.author = getAuthorName(auth.content.firstName, auth.content.lastName) );
-      author = promise_author.author
+      // author_obj.then( auth => promise_author.author = getAuthorName(auth.content.firstName, auth.content.lastName) );
+      // author = promise_author.author
+      author_obj.then( auth => console.log(`Resolved Author: ${auth}`) );
+      author = "Placeholder Author Name";
     } else {
       author = getAuthorName(author_obj.content.firstName, author_obj.content.lastName);
     };
--- a/apps/consulting/utils/getLibraryTiles/getBlogArticlesProps.ts
+++ b/apps/consulting/utils/getLibraryTiles/getBlogArticlesProps.ts
@@ -11,27 +11,6 @@ export const getBlogArticlesProps = (blogArticles: PageItems): TTiles =>
         (bodyItem) => bodyItem.component === 'blog-article',
       ),
     };
-
-    if ( typeof blogArticleBodyItem.author.content === 'undefined' ) {
-      console.log(`Title: ${blogArticleBodyItem.postTitle}`);
-      console.log(`Incoming Author: ${blogArticleBodyItem.author}`);
-    }
-
-    // blogArticleBodyItem.author could be a Promise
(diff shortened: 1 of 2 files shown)


bskinn:
All three `consulting` builds have built cleanly on three consecutive deploy sets. I think this may be where the problem has been -- a stale/mangled cache.

@marcin-antczak-naturaily / @kherma, would there be any problems associated with this change to the `network-only` policy for this one specific query?

marcin-antczak-naturaily (approved):

PR #356, opened 2022-07-18
refactor: One Vercel project per website

gabalafou:
Companion [proposal PR](https://github.com/Quansight/Quansight-website/pull/357).

This pull request attempts to simplify our current Vercel integration with
Storyblok and this code repo. We currently have three projects in Vercel per
website (for a total of six projects in Vercel). The cognitive overhead of this
setup is too high.

The projects in Vercel result partly from the number of layers involved in
generating the website. Each of these layers has variables. Stacking the layers
and their variables together produces many variants of the website. When the
site is built, it combines Storyblok content with GitHub code in a particular
Vercel environment. Storyblok has published versus saved content. Our GitHub
repo has develop versus main branch code. Vercel has preview versus production
environments. This is summarized in the following table:

| Storyblok          | GitHub          | Vercel                |
| ------------------ | --------------- | --------------------- |
| [saved, published] | [develop, main] | [preview, production] |

Multiplying these all together produces 8 variants.

The main trick of this pull request is simply eliminating some of these possible
combinations. For example, it is not considered necessary to combine unpublished
content with main-branch code because whatever is in the develop branch should
eventually make its way to main, so it would be better to look at
unpublished content against develop-branch code; that way any potential errors
resulting in the combination of content plus code have a better chance of being
caught sooner. This also reduces the number of combinations for reviewers to
review.

In short, the only combinations permitted after this PR are:

```
preview, develop, saved
preview, develop, published
production, main, published
```

Technically there is a third pair of combinations to support local development,
but I have omitted discussion of local development for brevity.

The other trick of this pull request is using the Next.js preview mode to vary
the site at run-time rather than at build-time. That way we don't have to have
two separate URLs for the product of: [Vercel preview] x [Storyblok saved,
Storyblok published]. We can present both views of the website at the same URL.
When Next.js preview mode is on, we show draft/saved content in
the Vercel preview environment. When preview mode is off, we show only published
content.

I have also tried to think holistically about the entire content workflow cycle.
This includes things like incorporating a Slack bot into the content workflow
process. It also includes rendering a thin banner at the top of every page of
the site when it's built in a Vercel preview environment. The banner tells you
whether you are seeing draft content, or published
content. It also tells you which code branch the site was built from. On the
production site, the banner is not rendered at all. On production, you can only
see published content against main-branch code.

## Summary of changes in this PR

- New shared component PreviewModeBanner, which displays a banner at the top of the website letting the end user know whether or not they are in Next.js preview mode.
- Add middleware.ts to Consulting and Labs in order to check server-side at request-time (i.e., even before the client-side code has loaded) if the request comes from the Storyblok visual editor. If so, then redirect the request through the /api/enter-preview endpoint to kick the user into preview mode. 
  - The reason I used middleware is that I couldn't find another way with Next.js to put the Storyblok visual editor into preview mode if the request was for an unpublished path (i.e., labs.quansight.org/i-am-a-new-unpublished-page). The redirect fallback config would match and handle the request before any of the React app client-side code had a chance to check whether the user was in the Storyblok editor and therefore bump them into preview mode in which the unpublished page would be picked up and handled by the app code rather than fallback to the old Labs site. See the part in the Next.js docs about [execution order](https://nextjs.org/docs/advanced-features/middleware#matching-paths).
- Middleware is a newer Next.js feature, so upgrade Next.js and the corresponding eslint plugin.
- Simplify/reduce environment variables
- Pass the Next.js `preview` boolean to all API helper functions in order to change the Storyblok-specific HTTP header `version` at run-time (in order to be able to switch between draft vs published content).
- api/preview.ts was serving the same purpose as api/enter-preview.ts. Consolidated the two into enter-preview.ts and deleted preview.ts.

## Migration Plan

The general idea is that the Vercel projects quansight-consulting-live and quansight-labs-live will become the only two Vercel projects (one Vercel project per website), and all of various non-localhost Storyblok visual editor URLs will be replaced with a single default URL. Also all of the tasks and webhooks in Storyblok will be deleted except for one webhook to deploy the `develop` git branch whenever somebody hits the publish button.

### Vercel

Before merging this code into the `develop` branch, the following changes need to be made in Vercel settings in order to test the changes and features in this PR. Fortunately, none of these changes are backwards incompatible.

Under Vercel, under quansight-labs-live and also under quansight-consulting-live, under Settings,

- Under General:
  - [x] In the section Build & Development Settings, Modify output directory to use dot-slash path: `/dist/apps/...` to `./dist/apps/...` (this surfaced from adding the middleware file; I reported it to Vercel)
  - [ ] In the section Project Name, remove the "-live" from the project name for each project (e.g., quansight-consulting-live becomes just quansight-consulting)
- Under Environment Variables,
    - Under production,
        - [x] Set `NEXT_PUBLIC_STORYBLOK_TOKEN` to the value of the **public** key in Storyblok. (Consulting and Labs each have their own separate key.) 
        - [ ] Clean-up tasks for later: delete `STORYBLOK_TOKEN_VERSION`, `STORYBLOK_TOKEN`, and `STORYBLOK_API_URL`
    - Under preview and also under development,
        - [x] Add/set `NEXT_PUBLIC_STORYBLOK_TOKEN` to the value of the **preview** key in Storyblok. (Consulting has separate key from Labs.)
- [x] Under Git, under the section Deploy Hooks, create a new deploy hook. Name it Storyblok Publish and use git branch `develop` (not `main`)

### Storyblok

Under Storyblok settings:

- [x] Change the **Story published & unpublished** webhook to the new one you created in the Vercel settings.
- [ ] Assuming you have changed the names of the Vercel projects, change **Location (default environment)** to https://quansight-consulting-develop-quansight.vercel.app/ (trailing slash is important) for consulting and https://quansight-labs-develop-quansight.vercel.app/ for labs.
- [ ] Clean up tasks for later: delete the tasks from Storyblok and all of the preview URLs except localhost:4200.

### Slack

- [ ] Add Vercel bot to #qwebsite-ops Slack channel.

Once the preview deploys of quansight-consulting-live and quansight-labs-live have been vetted and tested, this PR will be merged to develop, and I will immediately create a PR to merge `develop` into `main`. At that point, I will finish checking off any of the changes above that have not been checked off, and the whole system should be testable end-to-end. From there, we will merge to `main` and see if the production deploy works. If not, we will have to fix in real-time (so this should be done Monday), perhaps with hotfix PRs directly to the `main` branch. But if the production deploy works, then we will be in the new world, and after a day or two, we can go ahead and do the clean-up work of deleting projects, environment variables, preview URLs and other settings no longer needed.

## Other known to-dos:

- [x] Extend changes to Consulting site. Currently this PR only covers the Labs site.
- [x] Put the banner content in a summary/details tag in order to provide more info. Such as explaining why toggling out of preview mode in the Storyblok preview frame is disabled. 
- [x] Summarize changes both within this PR and to the systems outside of the PR (important for migration)

Pushing for after this PR:
- [ ] Improvements to the preview mode banner. Instead of a button, use a radio toggle? Rather than hide the button in Storyblok preview iframe, disable it. Maybe also provide a timestamp of when the page was statically generated.
- [ ] Fix preview mode banner in mobile so it doesn't interfere with the nav.
- [ ] When the git ref at build time is `develop`, default out of Next.js preview mode (this is covered in the [companion proposal PR](https://github.com/Quansight/Quansight-website/pull/357)).


gabalafou (commented):
Done self-reviewing

gabalafou:
@tokordys I don't know why I can't assign you to this PR

libs/shared/storyblok-sdk/src/hooks/useStoryblok.ts
```diff
@@ -11,7 +11,7 @@ interface CustomWindow extends Window {
 declare const window: CustomWindow;
 
 export const Storyblok = new StoryblokClient({
-  accessToken: process.env['NEXT_PUBLIC_STORYBLOK_TOKEN'],
+  accessToken: process.env['STORYBLOK_TOKEN'],
```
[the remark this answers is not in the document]
marcin-antczak-naturaily: https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

This variable should not be accessible without prefix on client side.
So, if we use this StoryblokClient instance to fetch data on client side we won't see an error because it's going to work without token too. Thing is that in this case we can only fetch data that is published an publicliy available.

package.json
```diff
@@ -32,7 +32,7 @@
     "gray-matter": "^4.0.3",
     "highlight.js": "^11.5.1",
     "lodash": "^4.17.21",
-    "next": "12.1.6",
+    "next": "12.2.2",
```
marcin-antczak-naturaily, line 35: version 12.2.5 is available already

marcin-antczak-naturaily (changes requested):

tsconfig.base.json
```diff
@@ -22,7 +22,10 @@
       "@quansight/shared/ui-components": [
         "libs/shared/ui-components/src/index.ts"
       ],
-      "@quansight/shared/utils": ["libs/shared/utils/src/index.ts"]
+      "@quansight/shared/utils": ["libs/shared/utils/src/index.ts"],
+      "@quansight/shared/utils/preview/middleware": [
```
gabalafou, line 26: See https://github.com/Quansight/Quansight-website/pull/356/files#r938741908

PR #391, opened 2022-07-25
[BLOG] Add array-libraries-interoperability.md

MarsBarLee:
## Text styling

## Non-text contents

no labs links found
no llc links found


gabalafou:
This blog post is a little challenging to transfer because it defines some inline styles. I don't think that's supported right now by our blog post compiler.

If you visit the blog post on the old site, you can find, for example, a spot where Ani styled a small section with a double column.

So this is going to take a bit more time and thought than I have at this exact moment.

gabalafou:
I replaced @anirrudh's side-by-side lists with a table that I think does a better job comparing and contrasting the pros and cons of the two libraries. But I will leave it to @anirrudh to make that call. Here's the [blog post preview](https://labs-di16c1vjk-quansight.vercel.app/blog/array-libraries-interoperability).

rgommers:
@gabalafou you've got the wrong Anirudh - it's @anirudhDagar's blog post. @anirudhDagar want to have a look at the above question?

AnirudhDagar:
Thanks a lot @gabalafou for updating the post and helping with the migration. It looks good to me with the table, the comparison is clear. Just one little thing, can we have Pros first and Cons later for both? Right now, the order is not symmetric. Hope we can migrate this post to the new site soon. :)

Thanks @rgommers for the tag.

gabalafou:
Oh sorry about that. @AnirudhDagar, are you sure you want the pros of uarray to be side-by-side with the pros of Array API? And ditto with the cons. That would be closer to the original way that it was presented; however, I noticed when I was making the table that it seems that many of the pros of uarray are the cons of Array API and vice versa. For example:

- Pro urray: Can handle compiled code (anything not Python but which ends up with Python bindings).
- Con Array API: Can't handle compiled code, works only for pure Python and array provider based code in consumer library methods.

But if you're sure you want it the way it was originally, I will make it so. 😌 Thanks!

AnirudhDagar:
Hi @gabalafou, I see the comparison, but some points are also standalone.
* uarray:  Supports the ability to coerce/convert inputs and wrapping other arrays using the __ua_convert__ protocol.
* Array API: Not all functions are covered in the Array API spec, which may be a blocker if the consumer library method utilizes a function outside Array API's scope.

 The idea was to list the features and limitations of using `uarray` and `Array API` side by side, but I can understand the comparison. I don't have a strong bias, but I'd prefer to mention the pros and cons like before. Thanks again for helping with this :)

AnirudhDagar:
Hey @gabalafou, I stumbled back on this PR today. Just putting a small reminder. Let me know if I can help here.
Thanks! :)

apps/labs/tailwind.config.js
```diff
@@ -3,7 +3,7 @@ const { join } = require('path');
 
 module.exports = {
   content: [
-    join(__dirname, '**/!(*.stories|*.spec).{ts,tsx,html}'),
+    join(__dirname, '**/!(*.stories|*.spec).{ts,tsx,html,md,mdx}'),
```
gabalafou, line 6: This change adds all `.md` and `.mdx` files to Tailwind CSS processing. I think all that means is that Tailwind will scan all of those files for class names so that it produces a CSS file that's as minimal as possible. But I'm not 100% sure, so we'll need to be sure to take an extra hard look at the Labs site.

There were alternatives to doing this. One possibility was to keep the `<style>`-tag styles that Anirudh had added to his blog post. But this made me a little nervous because I'm not entirely sure if Next.js removes them afterwards. Furthermore, Anirudh's styles didn't look quite as good in the new website because the text width was different, which caused the "cons" headings to be misaligned, depending on screen width. I also suspect that being able to use Tailwind styles in blog post MD/X files might be useful in the future, especially if we have more blog posts with special, one-off styling. 

@bskinn, I did not make this change Consulting's Tailwind configuration file, but perhaps we should for consistency? That is... once we establish that the change doesn't mess things up.

gabalafou:
Hey @AnirudhDagar, sorry about that! I honestly don't remember seeing a notification for your message on March 3. 

I just pushed some changes that [should return the formatting to the way you had it originally](https://labs-git-fork-marsbarlee-array-libraries-inter-ead1a3-quansight.vercel.app/blog/array-libraries-interoperability), please take a look

gabalafou:
Oh good catch. There's a discrepancy between my local dev environment and the Vercel preview environment. My local dev environment renders the sunshine emoji:



I will need to dig in and figure out where this discrepancy is coming from.

AnirudhDagar:
Thanks for looking in, but I'd say if it is taking too much effort to fix, it is probably not worth your time. In that case feel free to switch to text. I'll leave that to you! :)

apps/labs/posts/array-libraries-interoperability.md
```diff
@@ -0,0 +1,598 @@
...
+description: "Let's dive into the technicalities involved in array libraries interoperability and understand the protocols making this a reality."
+category: [Array API]
+featuredImage:
+  src: /posts/array-libraries-interoperability/feature.png
+  alt: 'A pro and cons list comparing two protocols, uarray and the Array API.'
+hero:
+  imageSrc: /posts/array-libraries-interoperability/hero.png
+  imageAlt:
+    'Meme of Master Splinter leading the baby turtles from TMNT. Splinter
+    represents NumPy, and the turtles represent TensorFlow, CuPy, PyTorch and JAX.'
```
AnirudhDagar, line 14: @gabalafou I just noticed that the featured image for the blog is this `feature.png`, which was not present initially. Instead, can we have the `ninja_arrays.png` as the feature image?

Sorry for making the requests one by one. I should have caught that before.

gabalafou: That seems like a reasonable request to me, but I didn't make that change or the feature.png image. There may have been some reason for it. Do you know, @MarsBarLee?

apps/labs/posts/array-libraries-interoperability.md
```diff
@@ -0,0 +1,598 @@
+---
+title: 'Array Libraries Interoperability'
+published: October 13, 2021
+author: anirudh-dagar
+description: "Let's dive into the technicalities involved in array libraries interoperability and understand the protocols making this a reality."
+category: [Array API]
+featuredImage:
+  src: /posts/array-libraries-interoperability/feature.png
```
MarsBarLee, line 8:
```suggestion
  src: /posts/array-libraries-interoperability/ninja_arrays.png
```
Change based on the author's request.

apps/labs/posts/array-libraries-interoperability.md
AnirudhDagar: Thanks @MarsBarLee! Yeah I like the image too, hence the request.😛

apps/labs/tailwind.config.js
pavithraes: @bskinn Pinging just to surface this comment for you. :)

pavithraes:
> I will need to dig in and figure out where this discrepancy is coming from.

FYI, I see the emojis in the deploy preview now: https://labs-git-fork-marsbarlee-array-libraries-inter-ead1a3-quansight.vercel.app/blog/array-libraries-interoperability#protocol_differences :)

apps/labs/posts/array-libraries-interoperability.md
```diff
@@ -0,0 +1,598 @@
+---
+title: 'Array Libraries Interoperability'
+published: October 13, 2021
+author: anirudh-dagar
+description: "Let's dive into the technicalities involved in array libraries interoperability and understand the protocols making this a reality."
+category: [Array API]
+featuredImage:
+  src: /posts/array-libraries-interoperability/ninja_arrays.png
+  alt: 'A pro and cons list comparing two protocols, uarray and the Array API.'
```
pavithraes, line 9:
```suggestion
  alt:
    'Meme of Master Splinter leading the baby turtles from TMNT. Splinter
    represents NumPy, and the turtles represent TensorFlow, CuPy, PyTorch and JAX.'
```

gabalafou (approved):

AnirudhDagar:
Thanks all for getting this migrated to the new blog site! :))

PR #413, opened 2022-08-11
Add labs.quansight.com -> .org redirect

### LINKED ISSUE #410: [FEAT] - Add redirect of `labs.quansight.com` to `labs.quansight.org`
### What site is this for?

Quansight LLC

### Expected behavior

_No response_

### Actual behavior

This issue serves to document what was done to achieve a redirect of `labs.quansight.com` to `labs.quansight.org`.

The first attempt was the NON CODE CHANGE of adding a CNAME entry to the Vercel DNS records for `quansight.com`:

- NAME: labs
- TYPE: CNAME
- VALUE: labs.quansight.org
- TTL: 3600
- PRIORITY: (empty)

This does not seem to have worked, as the `*` CNAME entry that Vercel uses to collect all incoming traffic to all subdomains appears to be taking precedence.

The solution instead may be a [Next.js redirect using `has.host`](https://stackoverflow.com/a/67211550/4376000) to select on the `labs` subdomain.

_No response_

### Anything else?

_No response_

bskinn:
Attempting to blanket-redirect all incoming paths to .org

Closes #410.


bskinn:
Need your help here, @gabalafou -- I tried to set up the redirect so that it would trigger even with the Vercel preview URLs, but it doesn't seem to be working right. When I prepend `labs.` to the preview URL, it complains of an SSL error.

I tried creating a static Vercel URL for the PR branch, https://quansight-redirect-test.vercel.app/, but the same thing still happens at https://labs.quansight-redirect-test.vercel.app/.

Do you have any thoughts on what's going on, and how to successfully test this before a merge to `develop` or `main`?

bskinn:
I think I got it. I had to create Vercel deployment domains on the `quansight-consulting-live` project at *both* https://quansight-redirect-test.vercel.app *and* https://labs-quansight-redirect-test.vercel.app. This provided a valid build of the site at a host starting with `labs` (satisfying the `^labs[.-].+` regex) to test the redirect, and one *not* starting with `labs` to serve the actual LLC site.

Please check those two links, @gabalafou, and see if you think this is a solid implementation.

If so, once the change is merged to `main`, we will need to create the `labs.quansight.com` domain for the `quansight-consulting-live` project, in order for the redirect to work on the live site.

bskinn:
The URL `:path` appears to be successfully transferred across the redirect, too; see e.g.:  https://labs-quansight-redirect-test.vercel.app/blog/the-evolution-of-the-scipy-developer-cli

noatamir:
Thanks Brian! LGTM, but I would prefer @gabalafou double checks.

apps/consulting/next.config.js
```diff
@@ -29,6 +29,17 @@ const nextConfig = {
         permanent: false,
         destination: '/careers',
       },
+      {
+        source: '/:path*',
+        has: [
+          {
+            type: 'host',
+            value: '^labs[.-].+$',
```
gabalafou, line 37: Could you add a line explaining this regex? For example, why are you matching "labs" at the beginning of the string, but then anything at the end?

gabalafou (approved):
A code comment for the regular expression would be nice, but otherwise, this looks good to me.

apps/consulting/next.config.js
bskinn: Sure, will do. The terminating `$` is not necessary, I'll remove for simplicity.

bskinn: Ok, comment added, thoughts pls @gabalafou.

gabalafou (approved):

apps/consulting/next.config.js
gabalafou: Looks good, thanks!

PR #415, opened 2022-08-12
[BLOG] Add feature image for Scipy 2022 Labs blog

bskinn:
The feature image was left unspecified, making the tile for the post look bad on the `/blog` page.

### DIFF
--- a/apps/labs/posts/alt-text-scipy-2022.md
+++ b/apps/labs/posts/alt-text-scipy-2022.md
@@ -4,9 +4,9 @@ author: tony-fast
 published: July 13, 2022
 description: "Announcing the SciPy 2022 Accessibility Awareness Efforts"
 category: [Access-centered]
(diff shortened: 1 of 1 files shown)


gabalafou:
I see what you mean, the lack of feature image is more pronounced with the addition of newer blog post:

gabalafou (approved):
This looks like a clear improvement to me, so approving. We can always change later if needed.

noatamir:
I wasn't a fan of that decision to begin with - thanks for the PR @bskinn!

bskinn:
@noatamir If it was a specific decision about that image, perhaps we should choose a different feature image?

bskinn:
(I had assumed it was missed in the rush to get the post published.)

noatamir:
Yeah, you are right, it was probably the rush. The image looks good, I think 🤔

PR #434, opened 2022-08-31
fix: update lock file

gabalafou:
This pull request is just the result of running `npm install`.

The lock file got out of date because I [accepted a suggestion to bump Next.js from 12.2.2 to 12.2.5](https://github.com/Quansight/Quansight-website/pull/356#discussion_r952432675) without running `npm install` afterwards.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "gray-matter": "^4.0.3",
         "highlight.js": "^11.5.1",
         "lodash": "^4.17.21",
-        "next": "12.2.2",
+        "next": "12.2.5",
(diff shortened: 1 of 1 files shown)


bskinn (approved):
So many JSONs.

As long as you're satisfied that the new deps are ok, I don't think I have anything useful to add on review.

bskinn:
Although, @gabalafou -- maybe we should take this opportunity to roll up all of the dependabot PRs?

gabalafou:
> Although, @gabalafou -- maybe we should take this opportunity to roll up all of the dependabot PRs?

Good idea. I'll make a separate PR for that.

PR #435, opened 2022-09-01
Delete CODEOWNERS

gabalafou:


bskinn:
Yeah, the PR template is probably as good a place as any ... It would put the info directly in the context where it's needed.

Might also be good to put it in README...or, to keep a single source of truth, to add to README a mention that recommended reviewer info is in the PR template.

bskinn (approved):
Let's get this CODEOWNERS deleted now. I've opened #446 as a place to discuss the details of how we want to define/document go-to reviewers for different situations.

PR #436, opened 2022-09-01
delete mergeable.yml

gabalafou:


bskinn:
@gabalafou Could change '1 approvals' to singular '1 approval'  in the error message for the approvals check

.github/mergeable.yml
```diff
@@ -1,30 +1,9 @@
...
-          title: 'Mergeable Run have been Completed!'
-
   - when: pull_request.*, pull_request_review.*
     name: 'Approval check'
     validate:
       - do: approvals
         min:
           count: 1
-          message: 'Pull request must have at least 1 approvals'
+          message: 'Pull request must have at least 1 approval'
```
trallard, line 9: I do not think we need this at all I think we can just delete the whole workflow since I have already set this rule in the actual repo settings. i.e. an admin needs to review and approve the PR before merge

bskinn (approved):
With approval to remove Mergeable entirely, LGTM.

Issue #452, opened 2022-09-14
[BUG] - Labs homepage template has a mismatch in paragraph flow between desktop and mobile/tablet


noatamir:
### What site is this for?

Quansight Labs

### Expected behavior

Mobile/tablet template should have the same order of paragraphs as the desktop template
<img width="1117" alt="desktop - correct order" src="https://user-images.githubusercontent.com/6564007/190224213-9c89c33b-080c-42ab-8296-21c106d441bd.png">

### Actual behavior

When we “solved” the text flow on the desktop, we rearranged the text by moving one paragraph content to another in the existing template. But on mobile they get transformed into a column of text, so they appear in the wrong order. We need to ask Gabriel to tweak the template in either mobile or desktop to resolve the issue. 

<img width="697" alt="Mobile - out of order" src="https://user-images.githubusercontent.com/6564007/190224082-6a30b003-4f0f-42dc-8b0e-6c4f8462eff5.png">

gabalafou:
To be honest, it's still unclear to me what the order of the paragraphs should be in a single column.

I think I understand what you're suggesting, but the actual words in the paragraphs when I read them suggest the following order:

1. The public benefit division ...
2. Sustaining the Future of Open Source
3. Projects like NumPy
4. We have had and continue to have
5. We also understand the needs of companies

The [Labs home page in Figma](https://www.figma.com/file/vyWM9qXdXCSAPlTfUjuqrL/Quansight-2021-Website?node-id=1181%3A616) suggests the same order to me (although it is a bit ambiguous).

When I dig into the [Quansight Storyblok Manual](https://docs.google.com/document/d/1HVaWKhepSxIQP1ZY9o7O_n3qCQpE8KdgD9KIaS4P-n4) and the [ColumnArticle.tsx](https://github.com/Quansight/Quansight-website/blob/develop/libs/shared/ui-components/src/ColumnArticle/ColumnArticle.tsx) code file, I wonder if something got lost in translation between the Figma design and its implementation in Storyblok and React. 

I think Naturaily may have thought that what we had was two rows, each with their own columns, but I think we really just have columns.

The problem as I see it is that when the Column Article template/blok is presented in two columns, the four sections that comprise the Column Article are presented in the following order:

A. Column 1:
  1. Left
  2. Final

B. Column 2:

  3. Right 
  4. Image

But when that same Column Article goes to a single column on mobile, you get the following order:

1. Left
2. Right
3. Final
4. Image

[...]

PR #455, opened 2022-09-14
Change section order of ColumnArticle at md breakpoint

gabalafou:
Short-term fix for #452.


gabalafou:
## Testing 

We need to examine all of the places that the `Column Article` block is being used on the websites. 

It's a shared component so it could be used on both Consulting and Labs. I'm not sure where or if it's being used on the Consulting site.

noatamir:
LGTM for Labs. Thanks for the quick fix!  🌟

bskinn (approved):
I checked, and currently ColumnArticie is **not used** anywhere I could find on `consulting`.

All OK from a LLC interaction standpoint.

I *did not review* from a code correctness standpoint.

trallard (approved):
LGTM - so approving

Issue #446, opened 2022-09-14
[MAINT] - Formalize recommended reviewers in PR template, with note in README


bskinn:
### What site is this for?

Quansight LLC, Quansight Labs

### Actual behavior

Per discussion in #435, we should indicate who the go-to reviewers are for the different types of PR the repo is likely to have.

At present, the plan is to put this information in the PR template, with an accompanying note in the README that indicates the PR template as the source of truth; plan is open for discussion.

Tasks:

- [ ] Identify the categories of PR
- [ ] Choose the go-to reviewer(s) for each category
- [ ] Revise the PR template
- [ ] Add note to README

trallard:
That is what the CODEOWNERS (now deleted file) is meant to help with identifying reviewers. 

For content - folks usually request reviews from Ralf, Noa, or myself - plus any other peers who they have worked with or would have any insights

For the broader codebase I anticipate it will be a mixture of Gabriel and whoever ends up in the review cycle for the LLC blog

bskinn:
&lt;checks the CODEOWNERS docs&gt;

Ah, looks like the CODEOWNERS was [under-configured](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) -- that's why it wasn't working in a useful way.

Based on your recommendations, @trallard, perhaps a new CODEOWNERS like the following would be good. If so, I can create a PR and we can refine there as needed?

```
# Gabriel as default owner
*   @gabalafou

# Noa as default reviewer for Labs blogs and assets
# Tania/Ralf/others to be manually added as appropriate
/apps/labs/posts   @noatamir
/apps/labs/public/posts   @noatamir

# Noa as default reviewer for changes to Labs Storyblok content
# Tania/Ralf/others to be manually added as appropriate
/apps/labs/CHANGELOG.md   @noatamir 

# Brian as default reviewer for changes to LLC Storyblok content
# Kate/others to be manually added as appropriate
/apps/consulting/CHANGELOG.md   @bskinn

```

trallard:
Can we turn the above in a PR for review @bskinn ?

bskinn:
Done, @trallard -- see #458

PR #458, opened 2022-09-19
Add CODEOWNERS

bskinn:
Per discussion in #446.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,8 +4,7 @@
 # Noa as reviewer for Markdown examples
 /examples   @noatamir
 
-# Noa as default reviewer for Labs blogs and assets
-# Tania/Ralf/others to be manually added as appropriate
+# Noa, Tania & Ralf as default reviewers for Labs blogs and assets
 /apps/labs/posts          @noatamir  @trallard  @rgommers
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,15 @@
+# Gabriel as default owner
+*   @gabalafou
+
```
trallard, line 3: Should also add /examples/ - Noa as default

.github/CODEOWNERS
```diff
@@ -0,0 +1,15 @@
...
+# Noa as default reviewer for Labs blogs and assets
+# Tania/Ralf/others to be manually added as appropriate
+/apps/labs/posts   @noatamir
+/apps/labs/public/posts   @noatamir
```
trallard, line 7: This should be the three of us - and whoever gets first to it will review
it is pretty much what we used to do on our prev site

trallard (changes requested):

.github/CODEOWNERS
bskinn: Added!

.github/CODEOWNERS
bskinn: Added, and adjusted spacing to align

trallard (approved):
LGTM 🚀 let's merge and take it from there

Issue #404, opened 2022-08-04
[FEAT] - Implement Google Ads on LLC site


bskinn:
### What site is this for?

Quansight LLC

### Actual behavior

The site needs to be configured to interop with our Google Ads campaigns. We have the needed HTML and AMP tag/script content to be inserted into the pages.

At this time, we wish a conversion event to be submitted to Google Ads when a visitor successfully submits their contact information into one of the `Form` Storyblok components on `/about-us`, on any one of the seven services pages, or on the `/qhub` page.

The `Form` component is implemented as a shared component, but it is currently only used on the LLC site, and so the conversion-event trigger can just go into that component. 

Likely, the setup scripts/tags to be injected into `<head>` should just go in the global template for the LLC site.

These details are open to discussion, if a better implementation is possible. Also, given the possibility that we will want to define additional conversion events in the future, the code modifications should be done in a way that streamlines addition of the code needed for these additional events.

bskinn:
From Brandon, as told to him by our Google Ads rep, as something (additional?) to accomplish as part of this issue:

> We will have to update our conversion tracking code on the website to setup enhanced conversions, to prevent engagement drops from 3rd party cookies going away. Please follow the 3 steps below.
> - Create a conversion tracking through Global Site Tag. Instructions [here](https://support.google.com/google-ads/answer/12216424?hl=en).
> - Accept customer data terms. Go to Tools & Settings --> Conversions --> Settings --> Customer Data Terms --> Accept & Save
> - Implement enhanced conversions through Global Site Tag. Go to the Conversions page on Google Ads and click into your new conversion action, scroll down to enhanced conversions, turn on, and follow implementation instructions. You can also view them [here](https://support.google.com/google-ads/answer/9888145?hl=en).

bskinn:
Our SEO/etc. consultants will be assisting with a redesign of our ads campaigns. Their technical requirements will need to be incorporated into the technical solution implemented here. More details to come once available.

tokordys:
hi @bskinn Did your consultants manage to define the technical requirements so we have a full picture of this ticket?

PR #503, opened 2022-10-12
Style/quan 15 inline code segment

kherma:


bskinn:
@noatamir @trallard Please review the Labs blog posts at the [build preview](https://labs-git-style-quan-15-inline-code-segment-quansight.vercel.app/blog), and comment on the new inline code formatting.

It looks pretty good to my eye, though the font weight might be a bit heavy.

The purple color also might look a bit too much like a visited link.

noatamir:
I did a couple of side by side comparisons on the CLI blog post. I think it's a bit too big and makes the lines break too early. 

side by side comparison for new inline code font, first example

side by side comparison for new inline code font, second example

kherma:
@noatamir @bskinn 
I thought that by "match to surrounding text" you mean to make is same size... because originally it was 85% of parent's text. So if inline code was in paragraph with font size 16px then the font size of the code is ~13.6px. Now inline code has same size as parent.

Screenshot 2022-10-12 at 18 34 28

kherma:
@bskinn We can use letter-spacing i think: https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing

bskinn:
@kherma Don't worry, what you did is great as a first pass. Now that we have an actual example to work with, though, we're seeing things we couldn't anticipate before.

kherma:
@bskinn There is also something like:[ font-kerning](https://developer.mozilla.org/en-US/docs/Web/CSS/font-kerning) I've never used that before but looks good. Okay, so what's next, should I wait for review and merge or you want to work on that one and test options?

trallard:
Re spacing - that is to be expected since the used typeface is a monospaced font where all characters occupy the same space. So modifying the kerning is definitely not something we should look at doing. 

Re weight - I checked in with Kamil and asked to bump to 500 but I have not reviewed the website (only a screenshot).
The main problem here is there is not enough contrast between the surrounding text and the code colour. This is a problem we have noticed elsewhere with our brand colours.
I will have to go back and review these changes again tomorrow when I am back online and propose adjustments

bskinn:
Per Tania, @kherma, please hold on further work with this PR until she provides her requested changes.

In the meantime, as appropriate, please move on to work on the other issues in the sprint. #404 seems the most natural next project to me.

Thanks!

trallard:
@kherma I just had a look at the font and styling - and think the best styling would look like this:

The_evolution_of_the_SciPy_developer_CLI___Labs

- #b2399a (replace the main branding colour with this plum variant which should be good even for folks overriding light with dark background)
- `font-weight: 400`
- sizing should be `95%` of the parent text

bskinn:
@trallard -- IIRC, you said this looks good to you, ready to merge?

trallard:
Yes this can be merged

bskinn (approved):

PR #519, opened 2022-10-19
[BLOG] Add accessibility-whos-responsible

MarsBarLee:
Link to [blog post, successful build](https://labs-git-accessibility-whos-responsible-quansight.vercel.app/blog/accessibility-whos-responsible).

## Text styling

## Non-text contents

### DIFF
--- a/apps/labs/posts/accessibility-whos-responsible.md
+++ b/apps/labs/posts/accessibility-whos-responsible.md
@@ -1,7 +1,7 @@
 ---
 title: "Accessibility: Who's Responsible?"
 published: March 25, 2021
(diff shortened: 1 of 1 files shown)


MarsBarLee:
Noa mentioned putting this PR on hold until it's decided how to handle category links.
Since it’s been a few months and the category links change has not been implemented yet, I think we could replace the link with Noa’s suggestion (`https://labs.quansight.org/blog?category=Access-centered&page=1`) and merge for now. The Access-Centered tag isn't the same as the jlaba11y but it works as an umbrella tag.

PR #534, opened 2022-10-26
[BLOG] Add spyder-terminal

MarsBarLee:
Link to [blog post, successful build](https://labs-git-spyder-terminal-quansight.vercel.app/blog/spyder-terminal).

## Text styling

## Non-text contents


MarsBarLee:
The feature here is from an image from the blog post but is not resized yet to 391x174 pixels. Should I resize it? Or will this be fine?

image

noatamir:
> Should I resize it? Or will this be fine?

Looks fine to me. Thanks for checking!

MarsBarLee:
@noatamir I've resized the image to 391x174 pixels and this is how it looks on my screen:
image

It seems a little cropped but it is clearer than the original, bigger resolution. Does it look good on your screen?

MarsBarLee:
@gabalafou @trallard I've updated this with Noa's suggestion. It looks ready to merge. Should I go forward with merging PRs that Noa already reviewed and I made small changes for?

gabalafou (changes requested):
Oh wait, actually I'm going to try to tweak the preview image because I think it looks bad

gabalafou:
This is how the feature image looks now.

When blog post is in the first position in the list:

feature image

When blog post is not in first position in the list:



On mobile:



Better don't you think?

MarsBarLee:
@gabalafou That does look better and clearer in those different contexts. Thanks!

gabalafou (approved):

PR #576, opened 2022-11-16
Release 22-11-16 czi blog fix

noatamir:
Merge `develop` to `main`.


apps/labs/posts/close-encounter-with-pandas-and-the-jedis-of-open-source.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: 'Close Encounter with pandas and the Jedis of open source'
-published: Novermber 14, 2022
+published: November 14, 2022
```
gabalafou, line 3: I'm not sure why GitHub shows this file being changed, must be a particular trait of GitHub that I don't fully understand. Because when I checked out your branch, `release-22-11-16`, locally and run `git diff main`, these changes to "close encounter with pandas" (which are actually already in `main`) [do not show up](https://gist.github.com/gabalafou/f5aa5b5eb450dcb37d0b99ca4f35489e), which is what I was expected.

gabalafou (approved):

apps/labs/posts/close-encounter-with-pandas-and-the-jedis-of-open-source.md
noatamir: Yeah same. That's what I saw when I opened the PR as well.

bskinn: I was playing around with this. I think it boils down to Github having two different `compare` algorithms.

E.g., if you do a two-dot compare, it *only* shows *actual text diffs* between the snapshots at the two points, like what `git diff` does.

If you do a three-dot compare, it looks at (something like) the *set math of commits* that are different between the two.

A great example is a `compare` between `main` and `develop`, which would occur in the context of opening a PR from `main` into `develop`.

Two-dot comparison (https://github.com/quansight/quansight-website/compare/main..develop) is the same as what you get from a local `git diff` and shows no differences:

image


Three-dot comparison (https://github.com/Quansight/Quansight-website/compare/develop...main), which is the default, and which Github uses when creating a PR, *does*:

image

As best I can tell, in this case it shows stuff because of all of the merge commits on `main` from release branches split from `develop`... these commits (naturally) aren't in `develop`'s history.

bskinn: Oh, that's really helpful -- thanks!

bskinn: I had no idea it was a *git* feature, either. I thought it was some "helpful" thing that GH had implemented.

noatamir: Thank you both! 🌷

PR #578, opened 2022-11-22
MAINT - Release 2022-11-22

trallard:
Changes added in `develop` since last release:

* feat(llc):[QUAN-12]: add global gtag script

* feat(llc):[QUAN-12]: report conversion on form submission

* feat(llc/labs):[QUAN-12]: extend Window interface with gtag property

* feat(llc):[QUAN-12]: pass window.location object to gtag_report_conversion after form validation

* Remove URL from gtag_report_conversion call

We don't want to navigate to a new page as a result of the conversion being reported. This was my (Brian's) mistake, misunderstanding the function of the URL here.

* feat(llc):[QUAN-12]: after cr

* Re-remove argument to gtag_report_conversion()

We don't want the page to navigate anywhere after
the form is submitted. This fix was accidentally reverted in the course of PR review.

* Add sourcing comment for gtag in _app.tsx

Co-authored-by: Brian Skinn <brian.skinn@gmail.com>

### DIFF
--- a/apps/consulting/index.d.ts
+++ b/apps/consulting/index.d.ts
@@ -1,6 +1,15 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-declare module '*.svg' {
-  const content: any
-  export const ReactComponent: any
-  export default content
+
+declare global {
+  interface Window {
+    gtag: any;
+  }
 }
+
+declare module '*.svg' {
+  const content: any;
+  export const ReactComponent: any;
+  export default content;
+}
+
+export {};
--- a/apps/consulting/pages/_app.tsx
+++ b/apps/consulting/pages/_app.tsx
@@ -12,6 +12,18 @@ function CustomApp({ Component, pageProps }: AppProps): JSX.Element {
       <Meta />
       <Component {...pageProps} />
 
+      {/* These two scripts are included verbatim as per gtag instructions */}
+      <Script
+        strategy="afterInteractive"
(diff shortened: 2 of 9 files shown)


gabalafou (approved):
Looks good, thanks!

That PR description makes it look like several different things happened in `develop` since the last release, but it's really just this one thing: 

- #554 

Do we need to update our privacy policy following this addition? I noticed that on my first attempt at testing the form, I was using Brave (shields up), and it blocked the request for the Google analytics script. It looks to me like the current privacy policy is vague enough (which I don't love) and wide enough to cover this use case. Just double checking.

I then tested on Chrome and it looks like things are in order. When I submit the form, I see a GET request to a Google server with `event=conversion` in the payload, and then the page displays a thank-you message.

I think Brian said that we won't be able to check on the Google side that things are working until this in production. Fingers crossed!

bskinn:
I've already reached out to Legal to request any updated language needed. We'll get that up ASAP.

Google's Tag Manager Assistant extension makes it look like everything is in order. Fingers crossed, yep, but I think it's functioning the way we need it.

trallard:
> That PR description makes it look like several different things happened in develop since the last release, but it's really just this one thing:

That is because I used the gh CLI to open the PR and it gets the commit history and adds it to the message itself

gabalafou:
In the meantime, should we close this PR without merging?

gabalafou:
I reverted #554 in #580

PR #594, opened 2022-12-06
[CONS BLOG] Acceleration in Python

bskinn:
## Text styling

## Non-text contents


ericdatakelly:
- [ ] Consider adding or removing citations to create consistency.  It's not clear why some are cited and some are not.  We can probably just cite the less commonly known ones.

ericdatakelly:
- [ ] The image is too small.  If the original, high resolution image is still available, we should use that in the post and save the other image as the featured image.  Irina might have an original.

ericdatakelly:
@bskinn this is ready for review.

bskinn:
> * The image is too small.  If the original, high resolution image is still available, we should use that in the post and save the other image as the featured image.  Irina might have an original.

@irinafumarel, do you have the original for [this image from the post](https://consulting-git-cons-blog-accel-in-python-quansight.vercel.app/posts/acceleration-in-python-which-is-right-for-your-project/pythonacc.png)?

irinafumarel:
@bskinn Here is a larger version of the diagram and also the AI file https://drive.google.com/drive/folders/1WHBG5VwbYJqxr2ABwMSJVpDyzKQoqul3?usp=sharing

bskinn:
Great, thanks @irinafumarel!

@ericdatakelly, can you swap in this larger version?

ericdatakelly:
Done.  Thanks @irinafumarel !

PR #597, opened 2022-12-07
[CONS BLOG] RAPIDS cuCIM: Porting skimage Code to GPU

bskinn:
## Text styling

## Non-text contents

### DIFF
--- a/apps/consulting/posts/performance-for-image-processing-with-cucim.md
+++ b/apps/consulting/posts/performance-for-image-processing-with-cucim.md
@@ -13,7 +13,7 @@ hero:
   imageAlt: 'Data visualization of Paris city'
 ---
 
-This post was co-written between Gregory R. Lee & Gigon Bae, (NVIDIA).
+This post was co-written between Gregory R. Lee & Gigon Bae (NVIDIA).
 
 ## Overview
 


ericdatakelly:
TODO: Contact RAPIDS and have them update their link to our post.

ericdatakelly:
> TODO: Contact RAPIDS and have them update their link to our post.

Or we could forward to the new url.

bskinn:
The URL shouldn't be changing -- if it is, we need to rename the .md file to match.

bskinn:
Weird, why would the link from their post have ever been incorrect?

ericdatakelly:
> Weird, why would the link from their post have ever been incorrect?

Not sure.  I remember testing those links when they went live.

bskinn:
I'll check Wix to see what URL the post originally went up under...it should live there.

We may need to add a redirect as you suggest, though, @ericdatakelly, since the post has been live at this URL for a few months now.

ericdatakelly:
Besides the url issue, this post is ready.  I went ahead and replaced some images with code snippets and a table because they were easy to do and help with accessibility.  The preview displays them well, so I think this works, but of course we can go back to the images if needed.

bskinn:
> I went ahead and replaced some images with code snippets and a table because they were easy to do and help with accessibility.

Great, I think this is what we want in general, for any of these posts. I can't recall for sure, but it's possible these were replaced with images because we didn't have a good code-block formatting option inside Storyblok and we needed _something_ to migrate.

Thanks!

bskinn:
Yeah, looking at a [Wayback snapshot](https://web.archive.org/web/20220526150150/https://www.quansight.com/blog), the URL should be what's linked from NVIDIA. I'll do the following:

- [ ] Rename this post to match the old URL
- [ ] Add a Vercel redirect from where the post has been living on the new site to the old URL

ericdatakelly:
If the path to this post changed, others might have changed too, and there are probably other sites/blogs that reference them.  We should check the other posts too.

bskinn:
> and a table because they were easy to do and help with accessibility

@ericdatakelly, do you know if it's possible to constrain the table width using only Markdown? This table only has two columns but is rendering the full width of the text column, and it doesn't look that good.

ericdatakelly:
> > and a table because they were easy to do and help with accessibility
> 
> @ericdatakelly, do you know if it's possible to constrain the table width using only Markdown? This table only has two columns but is rendering the full width of the text column, and it doesn't look that good.

I don't.  However, I think the platform is interpreting html and css, so it should be possible to use one of these tricks.  
https://stackoverflow.com/questions/36121672/set-table-column-width-via-markdown

Or are you going for pure markdown to avoid future conflicts?

apps/consulting/posts/performance-for-image-processing-with-cucim.md
```diff
@@ -0,0 +1,274 @@
...
+from CuPy’s on disk kernel cache in the case of a new Python session).
+
+The necessary changes typically involve:
+
+1. Use `cupy.asarray` to transfer array inputs to cuCIM functions from the host
+   to the GPU.
+2. Replace functions imported from CPU-based library with GPU equivalents as in
+   the following table:
+
+| CPU Module | GPU Module      |
```
bskinn, line 173: @noatamir / @gabalafou / @trallard -- do any of you know how we could constrain the max width of this table (say, to 50% of the story column width) via some sort of MDX syntax?

bskinn:
> I think the platform is interpreting html and css

The source is being interpreted as MDX (a JSX/MD hybrid), and I'm still not very good at making it work. Left a comment inline requesting assistance.

apps/consulting/posts/performance-for-image-processing-with-cucim.md
bskinn: Never mind, adding the table as raw, Tailwind-styled HTML worked well, once I added `.md` and `.mdx` to the Tailwind content search config.

apps/consulting/posts/performance-for-image-processing-with-cucim.md
```diff
@@ -0,0 +1,304 @@
...
+from CuPy’s on disk kernel cache in the case of a new Python session).
+
+The necessary changes typically involve:
+
+1. Use `cupy.asarray` to transfer array inputs to cuCIM functions from the host
+   to the GPU.
+2. Replace functions imported from CPU-based library with GPU equivalents as in
+   the following table:
+
+<div class="flex md:justify-center md:pl-0 justify-left pl-16">
```
bskinn, line 173: Not the most elegant solution in-source, @ericdatakelly, but I think I got the table looking pretty good. See what you think in the [preview](https://consulting-pgp0li5a8-quansight.vercel.app/post/performance-for-image-processing-with-cucim#adding-gpu-support-to-existing-cpu-based-code).

ericdatakelly: It looks great!

PR #606, opened 2022-12-07
[CONS BLOG] Will Python Be #1 Forever?

bskinn:
## Text styling

## Non-text contents


ericdatakelly:
The plot doesn't match the text in the second paragraph.  He refers to Cobol, for example, but the plot does not contain Cobol.  Should we fix this or retain the original text?  One option is to add a second plot (adding too many languages to the first plot makes it too busy).

bskinn:
@ericdatakelly This effort is *just* for migrating the posts in their existing state. Please note in the comment thread here things that need changes, as you've done, and we'll create new content issues out of these PRs for downstream discussion and possible action.

apps/consulting/posts/will-python-be-1-forever.md
```diff
@@ -0,0 +1,35 @@
...
+can maintain the premier position in language popularity indefinitely. I’m sure
+there were many who believed that Java would never be replaced as #1.
+
+![](/posts/will-python-be-1-forever/python-1-img-1.jpg)
+
+_([Image source: PYPL][pypl page])_
+
+But that certainly does not mean that Python is going away completely. Virtually
+none of the languages that have come to prominence in the past have disappeared
+yet. They continue on as legacy code. Looking at the bottom right of the same
```
ericdatakelly, line 29: The text leads the reader through the plot, but the languages described in the text are not present in the plot.  Consider replacing the plot, adding a plot, or modifying the text.

ericdatakelly: That looks a lot better.  Thanks!

PR #613, opened 2022-12-14
[FEAT]: Add pass-through of UTM parameters to Zapier

bskinn:
In order to provide data on the performance of digital marketing campaigns, it's necessary to capture UTM tags on incoming traffic, and store them with the Leads in Pipedrive.

This capture should enable better tracking of when/how marketing campaigns lead to, e.g.:

- Contact form submissions (early-in-pipe/prospect conversions)
- Closed deals (end-of-pipe/client conversions)

If we decide it's necessary, we can also add capture of Google's GCLID, which would enable Offline Conversion Tracking for search ads optimization. GCLID capture is disabled in this PR, however.

---

This PR inserts a script into the header that captures these tags from incoming traffic and stores them in `sessionStorage`. The storage keys on `sessionStorage` are prefixed with `quansight_` to minimize any chance of name collision with other sites that might be performing similar capture.

image

image

Upon contact form submission, the parameters stored in `sessionStorage` are retrieved and bundled with the form-data payload for submission to the Zapier webhook.

image

If we decide to purchase ads on other platforms that provide similar conversion-tracking/ads-optimization functionality, we may want to add those parameters now or in the future. Otherwise, UTM codes will suffice.

---

Once this code is active on the site, the Zaps will need to be adjusted to pass the parameters into the Leads they create; and then either (a) existing fields in Pipedrive will need to be chosen for storing these parameters, or (b) new fields will need to be created for them in the Pipedrive schema.

### DIFF
--- a/apps/consulting/pages/_document.tsx
+++ b/apps/consulting/pages/_document.tsx
@@ -1,19 +1,31 @@
 import { FC } from 'react';
 
 import { Html, Head, Main, NextScript } from 'next/document';
+import Script from 'next/script';
 
 const Document: FC = () => {
   return (
     <Html lang="en" className="text-[62.5%]">
       <Head>
+        <Script id="session-store-url-params" strategy="beforeInteractive">
+          {`
+            const urlSearchParams = new URLSearchParams(window.location.search);
+            const params = Object.fromEntries(urlSearchParams.entries());
+            for (const param in params) {
+              if (
+                param.toLowerCase().startsWith('gclid') || 
+                param.toLowerCase().startsWith('utm_')
+                ) {
+                  window.sessionStorage.setItem(param, params[param])
+                }
+            }
+          `}
+        </Script>
         <link
           href="https://fonts.googleapis.com/css2?family=Inter:wght@200;400;600;700;800&family=Mukta:wght@400;500;700;800&display=swap"
           rel="stylesheet"
         />
-        <link
-          rel="shortcut icon"
-          href="/favicon.png"
-        />
+        <link rel="shortcut icon" href="/favicon.png" />
       </Head>
       <body>
         <Main />
--- a/apps/consulting/pages/_document.tsx
+++ b/apps/consulting/pages/_document.tsx
@@ -7,15 +7,25 @@ const Document: FC = () => {
   return (
     <Html lang="en" className="text-[62.5%]">
       <Head>
+        {/*
+          This script collects any GCLID and UTM query parameters from incoming
+          traffic and stores them in `sessionStorage`. It uses a 'quansight_'
+          prefix on the names of all stored parameters to reduce the possibility
+          of collision with other sites that might be using sessionStorage for
+          similar purposes.
+
+          Search params approach from https://stackoverflow.com/a/901144/4376000.
+        */}
         <Script id="session-store-url-params" strategy="beforeInteractive">
           {`
             const urlSearchParams = new URLSearchParams(window.location.search);
             const params = Object.fromEntries(urlSearchParams.entries());
             for (const param in params) {
               if (
-                param.toLowerCase().startsWith('gclid') || 
+                param.toLowerCase().startsWith('gclid') ||
                 param.toLowerCase().startsWith('utm_')
                 ) {
+                  // Prefix with 'quansight_' to minimize possible name collisions
                   window.sessionStorage.setItem('quansight_' + param, params[param])
                 }
             }
--- a/apps/consulting/pages/_document.tsx
+++ b/apps/consulting/pages/_document.tsx
@@ -15,6 +15,10 @@ const Document: FC = () => {
           similar purposes.
 
           Search params approach from https://stackoverflow.com/a/901144/4376000.
+          
+          'afterInteractive' strategy chosen here to minimize any reduction of
+          page load speed. It seems unlikely that anything could happen to disrupt
+          the UTM parameters in the `window.location` prior to their capture here.
         */}
         <Script id="session-store-url-params" strategy="afterInteractive">
           {`
--- a/apps/consulting/pages/_document.tsx
+++ b/apps/consulting/pages/_document.tsx
@@ -23,14 +23,13 @@ const Document: FC = () => {
         <Script id="session-store-url-params" strategy="beforeInteractive">
           {`
             const urlSearchParams = new URLSearchParams(window.location.search);
-            const params = Object.fromEntries(urlSearchParams.entries());
-            for (const param in params) {
+            for (const [key, value] of urlSearchParams) {
               if (
                 // param.toLowerCase().startsWith('gclid') ||
                 param.toLowerCase().startsWith('utm_')
                 ) {
                   // Prefix with 'quansight_' to minimize possible name collisions
-                  window.sessionStorage.setItem('quansight_' + param, params[param])
+                  window.sessionStorage.setItem('quansight_' + key, value)
                 }
             }
           `}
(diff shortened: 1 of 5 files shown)


apps/consulting/pages/_document.tsx
```diff
@@ -1,19 +1,45 @@
...
+          extensively we will be running Google Ads, and it simplifies any
+          privacy questions to omit its capture.
+
+          Search params approach from https://stackoverflow.com/a/901144/4376000.
+        */}
+        <Script id="session-store-url-params" strategy="beforeInteractive">
+          {`
+            const urlSearchParams = new URLSearchParams(window.location.search);
+            const params = Object.fromEntries(urlSearchParams.entries());
+            for (const param in params) {
```
gabalafou, line 27: I think you can do:

```suggestion
            for (const [key, value] of urlSearchParams) {
```

apps/consulting/pages/_document.tsx
```diff
@@ -1,19 +1,45 @@
...
+              if (
+                // param.toLowerCase().startsWith('gclid') ||
+                param.toLowerCase().startsWith('utm_')
+                ) {
+                  // Prefix with 'quansight_' to minimize possible name collisions
+                  window.sessionStorage.setItem('quansight_' + param, params[param])
```
gabalafou, line 33:
```suggestion
                  window.sessionStorage.setItem('quansight_' + key, value)
```

apps/consulting/pages/_document.tsx
```diff
@@ -1,19 +1,45 @@
...
+          of collision with other sites that might be using sessionStorage for
+          similar purposes.
+
+          The GCLID capture is disabled for now, since we're not sure how
+          extensively we will be running Google Ads, and it simplifies any
+          privacy questions to omit its capture.
+
+          Search params approach from https://stackoverflow.com/a/901144/4376000.
+        */}
+        <Script id="session-store-url-params" strategy="beforeInteractive">
```
gabalafou, line 23: Not that it probably makes any discernible difference, but is `beforeInteractive` really the right value for strategy here?

libs/shared/ui-components/src/Form/Form.tsx
```diff
@@ -35,7 +40,13 @@ export const Form: FC<TFormProps> = (props) => {
   };
 
   const onSubmit = handleSubmit((formValues): void => {
-    sendFormData(hookUrl, formValues)
+    const paramValues: ParamValues = getParamValues();
+    const combinedValues: FormAndParamValues = combineFormParamValues(
+      formValues,
+      paramValues,
+    );
```
gabalafou, line 47: Why not do the following?

```suggestion
    const combinedValues = {...formValues, ...paramValues}
```

libs/shared/ui-components/src/Form/utils.ts
```diff
@@ -12,3 +18,50 @@ export const getFormHeader = (props: TFormProps, state: FormStates): string => {
...
   }
 };
+
+/**
+ * Retrieves any GCLID & UTM parameters saved to sessionStorage.
+ * GCLID capture disabled for now.
+ *
+ * @returns {ParamValues} - Encapsulated GCLID/UTM parameters
+ */
+export const getParamValues = (): ParamValues => {
```
gabalafou, line 28: I know naming is hard, but "param values" is really generic, hard to search for. `getTrackingParams`?

gabalafou (changes requested):
Looking good, I have a few questions, maybe some requested changes, so I'm marking this "request changes" until those have been resolved.

Also, I'm not sure it's worth moving the code around but it looks like we're putting LLC-specific code in the folder shared by both LLC and Labs.

apps/consulting/pages/_document.tsx
```diff
@@ -1,19 +1,45 @@
...
+          {`
+            const urlSearchParams = new URLSearchParams(window.location.search);
+            const params = Object.fromEntries(urlSearchParams.entries());
+            for (const param in params) {
+              if (
+                // param.toLowerCase().startsWith('gclid') ||
+                param.toLowerCase().startsWith('utm_')
+                ) {
+                  // Prefix with 'quansight_' to minimize possible name collisions
```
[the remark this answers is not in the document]
bskinn: Oh, is it? I don't know one way or the other, not a web expert. If it is domain-specific, then yeah, prefix not necessary.

apps/consulting/pages/_document.tsx
```diff
@@ -1,19 +1,45 @@
...
+        {/*
+          This script collects any GCLID and UTM query parameters from incoming
+          traffic and stores them in `sessionStorage`. It uses a 'quansight_'
+          prefix on the names of all stored parameters to reduce the possibility
+          of collision with other sites that might be using sessionStorage for
+          similar purposes.
+
+          The GCLID capture is disabled for now, since we're not sure how
+          extensively we will be running Google Ads, and it simplifies any
+          privacy questions to omit its capture.
```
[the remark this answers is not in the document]
bskinn: :+1:, will relocate

apps/consulting/pages/_document.tsx
bskinn: I debated about that. `afterInteractive` should be fine, too? It should be a super-quick-to-run script, so I figured pushing it as early as possible would be fine, not slow pageload much. Totally fine to switch it, though.

bskinn:
> Also, I'm not sure it's worth moving the code around but it looks like we're putting LLC-specific code in the folder shared by both LLC and Labs.

The `Form` component lives in `/lib/shared`, so I don't know how to avoid this without a refactor.  Labs site doesn't use `Form` currently (and AFAIK doesn't plan to?), so we could probably relocate it to `/apps/consulting`. Not sure how high a priority that is relative to other tasks.

libs/shared/ui-components/src/Form/utils.ts
bskinn: V good point, will change.

libs/shared/ui-components/src/Form/Form.tsx
bskinn: Showing off my novice JS skills here. Thx.

apps/consulting/pages/_document.tsx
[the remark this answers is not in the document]
bskinn: [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage):

> The read-only sessionStorage property accesses a session [Storage](https://developer.mozilla.org/en-US/docs/Web/API/Storage) object for the current [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin).

[Also Mozilla docs](https://developer.mozilla.org/en-US/docs/Glossary/Origin):

> Web content's origin is defined by the scheme (protocol), hostname (domain), and port of the [URL](https://developer.mozilla.org/en-US/docs/Glossary/URL) used to access it. Two objects have the same origin only when the scheme, hostname, and port all match.

So, I think you're right, the prefix is unnecessary.

libs/shared/ui-components/src/Form/utils.ts
bskinn: Ok, I renamed everything -- see what you think

bskinn:
After thinking about this, it's probably best to keep the shared code in its shared. Labs might someday want a `Form`, and if they do, they may well want UTM codes passthrough implemented. They're fully anonymous, after all.

If they don't want UTM codes, then they can just implement the form without capturing/submitting them.

apps/consulting/pages/_document.tsx
gabalafou: I'm fine either way. Just thought it was worth asking about and perhaps also commenting why which strategy was chosen.

gabalafou (approved):
Didn't test but the code looks good to me 😄

PR #617, opened 2022-12-23
[FEAT] Add URL passthrough in Form component Zapier payload

bskinn:
Maintenance of the Zapier integration for the contact forms will be much simpler if we include identifying information for the page on which the contact form was filled out as part of the Zapier webhook payload. Currently, we have a separate Zap for each contact form on the website, which is a lot of duplication and extra work whenever changes are needed to the Zap configuration.

This PR adds code to inject `window.location.toString()` into the payload under the ~`url`~ `form_url` key. This allows the needed identification of the specific page source for a Lead to avoid the need for a separate Zap for each page. 

---

While it would be nice to also have a human-friendly identifier for the page as part of the payload (this would allow constructing a more user-friendly Zap), it's (a) not required, and (b) would require a great deal more work. So, this is a minimal PR, focused on the minimum of functionality.

If further refinement is needed here, options include:

1. Revising the `Form` component & Storyblok schema to include a `form_id` field, which would also be passed through to Zapier
2. Adding a mapping inside the site codebase that takes a URL and supplies a text description
3. Adding an automation step in Zapier to implement a similar mapping as (2)

(1) is probably the best option, as it stores the plaintext name of the page within the `Form` component where it is defined on that page.

Both (2) and (3) run the risk of desynchronization from the actual website sitemap. (3) has the additional disadvantage over (2) of requiring an additional Task within the Zap, and it may not even be practical to implement this sort of mapping lookup using Zapier tooling.

In theory, it might be possible to, e.g., introspect the `Page` that the `Form` component lives on, and extract the `title` text from the `Hero` (if it exists) to autogenerate a plaintext page name. But, it sounds like React is not really meant to be used in this way, and due to the fact that `Page` rendering pulls a dynamic set of Bloks as its `children`, likely any implementation here would have to live in Storyblok *anyways*, in which case (1) is going to be more straightforward.

### DIFF
--- a/libs/shared/utils/src/sendFormData/types.ts
+++ b/libs/shared/utils/src/sendFormData/types.ts
@@ -26,6 +26,7 @@ export type FormAndTrackingValues = {
   phone: string;
   company: string;
   message: string;
+  url: string;
   // gclid: string;
   utm_source: string;
   utm_medium: string;
--- a/libs/shared/utils/src/sendFormData/types.ts
+++ b/libs/shared/utils/src/sendFormData/types.ts
@@ -26,11 +26,18 @@ export type FormAndTrackingValues = {
   phone: string;
   company: string;
   message: string;
-  url: string;
   // gclid: string;
   utm_source: string;
   utm_medium: string;
   utm_campaign: string;
   utm_content: string;
   utm_term: string;
+  /*
+    We also include here the URL of the page where the contact form lives.
+    This will allow us to use a single, common Zap for all contact forms,
+    instead of having to define a separate Zap for each form.
+
+    See https://github.com/Quansight/Quansight-website/pull/617.
+  */
+  form_url: string;
 };
--- a/libs/shared/ui-components/src/Form/Form.tsx
+++ b/libs/shared/ui-components/src/Form/Form.tsx
@@ -44,7 +44,7 @@ export const Form: FC<TFormProps> = (props) => {
     const combinedValues: FormAndTrackingValues = {
       ...formValues,
       ...trackingParams,
-      url: window.location.toString(),
+      form_url: window.location.toString(),
     };
 
     sendFormData(hookUrl, combinedValues)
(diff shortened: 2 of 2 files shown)


libs/shared/utils/src/sendFormData/types.ts
```diff
@@ -26,6 +26,7 @@ export type FormAndTrackingValues = {
   phone: string;
   company: string;
   message: string;
+  url: string;
```
gabalafou, line 29: for clarity perhaps, we should call this `formUrl`? In this particular context it looks like it could be the contact's URL (like their company's homepage or something).

At the very least, I think we should put a comment here explaining what `url` is.

gabalafou:
> While it would be nice to also have a human-friendly identifier for the page as part of the payload (this would allow constructing a more user-friendly Zap)

How so? More user-friendly in what way? I haven't seen the other side of this integration in the Zapier admin, so I might be missing some context that would make what you wrote obvious.

gabalafou (approved):
Looks good to me, but could you change the variable name or add a comment? It's not critical but I think it would be good for code clarity.

bskinn:
> > While it would be nice to also have a human-friendly identifier for the page as part of the payload (this would allow constructing a more user-friendly Zap)
> 
> How so? More user-friendly in what way? I haven't seen the other side of this integration in the Zapier admin, so I might be missing some context that would make what you wrote obvious.

The final step in the Zapier automation is to send an email with the contact form payload, partly as (i) a direct-to-human notification of the new submission and partly as (ii) a data-redundancy backup. It would be nice to have a string like "Data Engineering & MLOps" seeded into the webhook inputs, for use in e.g. the subject line, in addition to `https://quansight.com/data-engineering-mlops`.

libs/shared/utils/src/sendFormData/types.ts
bskinn: Thanks for keeping me honest in name choices!

Changes made; pls re-review the new version, @gabalafou.

gabalafou (approved):

PR #619, opened 2023-01-02
[CONS BLOG] Evolving QHub to Nebari

bskinn:
## Text styling

## Non-text contents


apps/consulting/posts/evolving-qhub-to-nebari.md
```diff
@@ -0,0 +1,256 @@
...
+/>
+
+We’ve started hosting [bi-weekly meetings][nebari community meetings info] to
+give all community members, including users, contributors, and maintainers, a
+chance to collaborate and share ideas. We invite you to come by and say _hi_.
+
+Try out Nebari, and let us know how it goes. We look forward to seeing you in
+our community spaces!
+
+export const GoButton = () => {
```
bskinn, line 227: @gabalafou Can you help with this?

- This is probably really terrible React, cleanup welcomed
- For the life of me, I cannot find the Tailwind that will center this button in the blog text column
- I tried to import `Button` from `@quansight/shared/ui-components`, but it kept breaking the build
  - That said, I don't think we want exactly a `Button` here -- the padding on that shared componet is too wide on all four sides -- so this might not need to be solved
  - However, we likely do want to be able to import arbitrary components from the site codebase as a general capability, so figuring out how to do it here might be good

bskinn: Oh - I *had* it indented more nicely, but the auto-prettier in VS Code flattened it all to the left margin.

bskinn: Nevermind here, @gabalafou -- I figured out how to add new MDX-usable components (thanks for the README section there!), and I discovered that without .md and .mdx being defined in the Tailwind config, it wouldn't parse them for the classes used...and so it would minify out the styles I was trying to apply.

All fixed now.

PR #622, opened 2023-01-10
Release 2023-01-10

trallard:
Merge changes in `develop` as of `2023-01-10` to `main`

PR #621, opened 2023-01-10
[BLOG] Add post on Python packaging and workflows

rgommers:
## Text styling

- [x] The blog is written with [plain language](https://www.plainlanguage.gov/guidelines/) (where relevant).
- [x] If there are headers, they use the proper header tags in order to do so (with only one level-one header).
- [x] All links describe where they link to (for example, check the [Quansight labs website](https://labs.quansight.org/)).
- [x] Any kind of styling that the author uses (for example, bold for emphasis) is consistent throughout the blog.

## Non-text contents

- [x] All content is represented as text (for example, images need alt text and videos need captions or descriptive transcripts).
- [x] If there are emojis, there are not more than three in a row.
- [x] Don't use [flashing gifs or videos](https://www.w3.org/TR/UNDERSTANDING-WCAG20/seizure-does-not-violate.html).
- [x] If it were to be read as plain text, the blog still makes sense and no information is missing.

PR #623, opened 2023-01-10
Back-merge `release-2023-01-10` to `develop` to recombine the two development paths

bskinn:
Somehow, `develop` acquired a commit composed of the squash of `release-2023-01-10` after the point where `release-2023-01-10` branched off of `develop`.

This PR is a no-op to the actual filesystem contents of `develop`, but rejoins the dev path of `release-2023-01-10` into `develop`, to improve the cosmetics of the commit history.

I don't think this PR is *strictly necessary*, so if majority opinion is to close without merge I'm ok with that.


rgommers:
These commits were never in `develop`, so I don't understand how this can happen if the release process is merging `develop` into the production branch. But I'll leave the decision on this one to y'all, since I'm not really involved in website development & maintenance.

gabalafou (approved):

gabalafou:
For future reference for myself. The story (as far as I can piece it together from GitHub) goes like this:

- Ralf opened PR #621. It ended up with 7 commits.
- Ralf squash merged #621 as a single commit (02fc587) onto the 'develop' branch.
- Tania cut a release PR #622. This PR somehow had all of those 7 commits from #621 (that Ralf squashed into one).
- Brian merged #622 into 'main' (merge commit 3d4d53f).
- So now the commit histories of main and develop were different. The develop branch had a single commit (squashed) from Ralf, whereas the main branch 7 commits (plus the release merge commit). We don't know for sure how this happened exactly, but it did. Our best guess is that Tania cut the release branch from the feature branch associated with #621 instead of from 'develop', but that's just a guess.
- This PR essentially syncs the two branches so they will both have 8 commits: the 7 unsquashed commits plus the squashed commit.

PR #632, opened 2023-01-18
[FEAT] Add Video component 

galdranorn:


libs/shared/ui-components/src/Video/types.ts
```diff
@@ -0,0 +1,23 @@
...
+type TSpaceVariants = 'none' | 'small' | 'large';
+
+export type TVideoProps = {
+  url?: string;
+  thumbnail: TImage;
+  size: TSizeVariants;
+  align: TAlignVariants;
+  title: string;
+  sideText: TRichText;
+  text: TRichText;
```
gabalafou, line 14: Could we add some documentation as comments? Maybe give a full-fledged example of a video object? For example, what is `sidetext` versus `text` and why are both required? Are there constraints on topSpace and bottomSpace taken together, for example should topSpace === bottomSpace?

libs/shared/ui-components/src/Video/VideoThumbnail.tsx
```diff
@@ -0,0 +1,42 @@
...
+import Picture from '../Picture/Picture';
+import { TVideoThumbnailProps } from './types';
+
+const styles = {
+  button: `
+    absolute top-[50%] left-[50%] translate-x-[-50%] translate-y-[-50%] w-[6rem] h-[6rem] rounded-full z-1000 bg-violet transition-[.4s] group-hover:scale-[1.1]
+    before:content-[''] before:absolute before:top-[1.5rem] before:left-[2.2rem] before:w-0 before:h-0 before:border-t-[1.5rem] 
+    before:border-b-[1.5rem] before:border-l-[2.4rem] before:border-r-none before:border-transparent before:border-l-white
+  `,
+  wrapper: `
```
gabalafou, line 14: Never seen this pattern before. This is neat, I like it.

libs/shared/ui-components/src/Video/VideoThumbnail.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const wrapperStyles = clsx(styles.wrapper, hidden && 'hidden');
+
+  const handleClick = () => {
+    setHidden(true);
+    onClick();
+  };
+
+  return (
+    <div className={wrapperStyles} onClick={handleClick} aria-hidden={true}>
```
gabalafou, line 31:
```suggestion
    <div className={wrapperStyles} onClick={handleClick} aria-hidden={hidden}>
```

libs/shared/ui-components/src/Video/Video.tsx
```diff
@@ -0,0 +1,131 @@
...
+    header: `
+      mb-[5.6rem] text-[4rem] font-extrabold leading-[4.9rem] text-center font-heading text-violet
+      sm:mx-auto sm:max-w-[70rem] sm:text-[4.8rem] sm:leading-[5.3rem]
+      xl:mb-[6.8rem]
+    `,
+    player: clsx(
+      'relative pt-[56.25%] w-full',
+      videoSizeStyles[size || 'large'],
+    ),
+    text: clsx(textStyles, 'hidden', size === 'small' && 'lg:block'),
```
gabalafou, line 83: So is this always going to be hidden? Am I reading this right?

libs/shared/ui-components/src/Video/Video.tsx
```diff
@@ -0,0 +1,131 @@
...
+            className="absolute top-0 left-0"
+            url={url}
+            playing={playing}
+            config={playerConfig}
+          />
+        </div>
+        {sideText && (
+          <div
+            className={styles.text}
+            dangerouslySetInnerHTML={createMarkup(sideText)}
```
gabalafou, line 119: Just double checking: these values for `dangerouslySetInnerHTML` are only coming from Storyblok, right?

gabalafou (changes requested):
Several files are exact duplicates of each other. Can we dry/deduplicate this code, or is that even desired?

Marking this as "request changes" because I have questions.

apps/consulting/components/BlokProvider/mappers/getVideoProps.ts
```diff
@@ -0,0 +1,15 @@
+import { TVideoProps } from '@quansight/shared/ui-components';
+
+import { TVideoRawData } from '../../../types/storyblok/bloks/video';
```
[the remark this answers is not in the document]
bskinn: Because that's not the same location.

`@quansight` points to `/lib/shared`, while this relative import pulls from within `/apps/consulting`

libs/shared/ui-components/src/Video/types.ts
bskinn: Speaking from the perspective of the desired design: no, topSpace and bottomSpace should be independently adjustable.

bskinn:
> Several files are exact duplicates of each other. Can we dry/deduplicate this code, or is that even desired?

If the duplicated files all deal with Storyblok interactions, then I assume the duplication comes from the two separate Storyblok spaces for the two sites.

I'd be inclined against deduplication there, if it's even possible, since

1. The two schemas can potentially evolve independently
2. IIUC there's no common Storyblok code (except possibly for the core API calls, can't recall), so deduping would mean a significant refactor.

libs/shared/ui-components/src/Video/Video.tsx
galdranorn: It's `an optional rich text field that allows text to be placed in the open space to the right and left of the video, respectively` from the requirements so it will be visible if chosen size is 'small' on desktop view. So at the moment:
- if video size is other than 'small' - left/right settings are not applicable and side text is not showing,
- if video size is small - left/right settings are applicable on desktop and side text is showing on desktop - on the mobile there is full width view and side text is hidden.

Please confirm if this requirement was understood properly or we need to change something here.

libs/shared/ui-components/src/Video/Video.tsx
galdranorn: Yes, that's the way of implementing richText fields.

libs/shared/ui-components/src/Video/VideoThumbnail.tsx
galdranorn: OK, to be changed

libs/shared/ui-components/src/Video/Video.tsx
Katebrack: Sounds good @bskinn

libs/shared/ui-components/src/Video/VideoThumbnail.tsx
galdranorn: @gabalafou your proposal is causing [this kind of error](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/93f78856655696a55309440593e0948c6fb96134/docs/rules/click-events-have-key-events.md), so I think we should keep aria-hidden true anyways. changed syntax to proposed by a11y: `aria-hidden="true"`.

libs/shared/ui-components/src/Video/types.ts
gabalafou: I feel like my overall review comment got lost in the bit about constraints. That was meant to be an example to show that I think there needs to be a code comment (with examples) in the codebase for someone who has to later debug this code.

I'm asking for something like this:

```typescript
/**
 * - url: the URL where the video lives
 * - thumbnail: an object containing the alt text plus the name of the jpeg file
 *   in Storyblok that is a thumbnail for the video
 * - size: ?
 * - align: ?
 * - title: the title of the video
 * - sideText: ?
 * - text: ?
 * - topSpace: ?
 * - bottomSpace ?
 *
 * Are all combinations of possible values allowed? What constraints or other
 * information about this datatype might be useful to know here?
 *
 * One or two examples:
 *
 *   {
 *     url: '',
 *     thumbnail: {
 *     }
 *     ...
 *   }
 *   ...
 */
export type TVideoProps = {
```

libs/shared/ui-components/src/Video/VideoThumbnail.tsx
gabalafou: That would be bad for accessibility. Here's why. I applied this PR. Then I loaded a sample video on a page. Then I visited the page using a screen reader (VoiceOver on Mac). I was unable to find and play the video using only the screen reader. Which makes sense. If you apply `aria-hidden="true"` to an element, you're telling the browser to hide that element from the screen reader. 

There are a couple other accessibility issues:

- The play button needs a label. A person who cannot see has no way of knowing that the button is a play button. The screen reader just says "button."
- There's no way to stop the video. People need to be able to push pause after pushing play.

gabalafou (changes requested):
Sorry for the delay in responding about aria-hidden. I didn't have time when I last reviewed this to dig into the subject, and then it got pushed by other priorities.

Katebrack:
Hi @galdranorn @kherma 

In addition to the accessibility items identified by @gabalafou here is our feedback for the component with screenshots. Thank you! cc: @bskinn 

**All Devices**
At all video sizes, the “top space” and “bottom space” appear to be broken and have no effect. 

image6

image3

**Large Video View**
The text below the video should be aligned-center but left-justified to avoid white space on the right (see below image). 

image4

- Let’s make the text block stretch across 80% of the video so there’s a bit of white space on either side. Also aligned centered and justified-left (for desktop only). 

- New addition: In addition to the headline/title already implemented above the video, we’d also like subheadlines to be implemented both above and below the video. 

- The subheadline above the video will be placed immediately below the headline/title for all screen sizes. 

- The subheads are new and the font will be black (not purple) and the font size ~3 font sizes larger than body copy, with text both aligned center and center-justified

**Tablet/mobile**
When the video size is small and video align is left or right, the side text in tablet and mobile should still render above or below the video—above the video when video align is set to ‘right’ , below the video when video align is set to ‘left’. Instead the side text is going away entirely instead of being rendered above/below the video as requested.

image1

image7

image2

**Desktop**
The small video ‘center’ position setting  is not working at all…still showing the side text and video is not centered.
In this mode, the side text should NOT be shown on any screen size
When the video size is medium (on desktop) the video is not aligned center.

image5

When the video is at medium, the text should be aligned-center, left justified and same width as the video 100%.

galdranorn:
@gabalafou CC @bskinn 
According to your accessibility concerns: my minimum plan is to add the button labels and implement way (probably a button) to stop the video. But I wanted to make a slightly wider research for React Player accessibility options or maybe more accessible friendly alternatives (ex. AblePlayer recommended by HolisticA11Y for React) if switching plugin won’t we a time consuming matter and can provide additional functionalities. I plan to test it with VoiceOver after implementation to check “organoleptically” if it meets requirements.

marcin-antczak-naturaily:
@bskinn Hi, I've fixed functional issues. Controls on embedded youtube or vimeo objects are embedded. So, I don't see a way to change their setting.

bskinn:
Ok, thank you very much, @marcin-antczak-naturaily!

@Katebrack, we'll need to do another round of functional review here.

@gabalafou, do you have any ideas for reaching into the embedded YT/Vimeo widgets to achieve the accessibility on play/pause controls?

PR #517, opened 2022-10-18
[BLOG] Add putting-out-the-fire

MarsBarLee:
Link to [blog post, successful build](https://labs-git-putting-out-the-fire-quansight.vercel.app/blog/putting-out-the-fire).

## Links to other Quansight blogs:
There is no need to check for Blog2Blog links due to updated website build. 
## Text styling

- [x] The blog is written with [plain language](https://www.plainlanguage.gov/guidelines/) (where relevant).
- [x] If there are headers, they use the proper header tags in order to do so (with only one level-one header).
- [x] All links describe where they link to (for example, check the [Quansight labs website](https://labs.quansight.org/)).
- [x] Any kind of styling that the author uses (for example, bold for emphasis) is consistent throughout the blog.

## Non-text contents

- [x] All content is represented as text (for example, images need alt text and videos need captions or descriptive transcripts).
- [x] If there are emojis, there are not more than three in a row.
- [x] Don't use [flashing gifs or videos](https://www.w3.org/TR/UNDERSTANDING-WCAG20/seizure-does-not-violate.html).
- [x] If it were to be read as plain text, the blog still makes sense and no information is missing.

PR #637, opened 2023-01-26
[BLOG] Update putting-out-the-fire images

isabela-pf:
This PR provides updated images to the blog migration at #517 per request. More details on the blog migration can be found at #517.

cc: @noatamir and @MarsBarLee


noatamir:
Thanks @isabela-pf!

I am seeing white edge next to this hero in the preview, which is breaking our design:
Screenshot of the preview with red oval highlights the white edge on either side of the hero image

Is it possible to change this image to the size/proportions of the hero template? (that would be 1440x696 pixels)

isabela-pf:
Hi there! Sorry for losing this thread, but I double checked and the image is the appropriate 1440x696 pixels. 

If there's something else wrong with the image please let me know. Also, if it's preferred to use the default image because the dark background doesn't meet the standards, I am fine with that.

pavithraes (approved):
Thank you both for verifying!

PR #645, opened 2023-02-01
Annual report pdf tracking

gabalafou:
This PR creates the following URL:

- `labs.quansight.org/annual-reports/quansight-labs-annual-report-2022.pdf`

This URL route is not just a simple file download; it uses Next.js Edge API Routes to:

1. first send tracking data to Plausible,
2. then return the PDF file to the user.

For marketing purposes the URL works just like any other URL. If you want to attach UTM data to the URL, you can do it like this:

- `labs.quansight.org/annual-reports/quansight-labs-annual-report-2022.pdf?utm_source=EthicalAds&utm_campaign=AnnualReport`

Those UTM params will get forwarded to Plausible when the user clicks that link. There is no redirect, by the way—the user clicks the link, Vercel sends a tracking event to Plausible and then serves the user the PDF. If the user were to copy-paste the URL above into their browser, they won't see it change in their address bar. They won't see the request to Plausible (because it happens server side). They'll just see that they entered a URL and got back a PDF.

As far as I can tell, we get all of the same data using this method that we would get with an HTML page EXCEPT for what kind of device the user was using (mobile vs tablet vs desktop). That's because Plausible uses screen size as the heuristic for device type, and the browser doesn't send that info automatically to the server. A normal HTML page on our website, on the other hand, runs the Plausible script, which asks the browser for the screen size and then sends it to Plausible's servers. But so long as we can do without device type info, this approach should be good to go.

For the record, I also considered using a normal URL route—labs.quansight.org/annual-reports/annual-report-2022 (not pdf)—that sends an HTML page to the end user that then automatically redirects to the PDF file. I thought that would be the simplest and most straightforward approach because we could fully leverage the Plausible script. Furthermore, when we eventually do put an HTML landing page at that URL that links rather than redirects to the PDF, we would have continuity across our analytics data (it would be the same `pageview` in Plausible both before and after). And we won't have to go back and change any of the ads or links. However, in order to give the Plausible script enough time to send data, I would have to put a delay on the redirect, and I discovered that [adding a delay to the redirect is an accessibility anti-pattern](https://www.w3.org/WAI/WCAG21/Understanding/change-on-request). There might be ways around it, but it would require me to test, and so this PR seemed like a faster implementation path. Furthermore, I also think it is somewhat of an anti-pattern to have a normal URL redirect to a PDF file, or at the very least, I don't think it represents the highest form of web citizenship, which is what I like to think that Quansight aims for. Which leaves us with one option: a mostly empty landing page that has a _Download PDF_ button that the user can click. But that's also not an ideal user experience in my humble opinion (because it requires two clicks instead of one to get to the content that user wants). The best user experience, as I thought about it, repeatedly took me back to what I have implemented in this PR. The trade-off is a bit more work on our end.

### DIFF
--- a/apps/labs/next.config.js
+++ b/apps/labs/next.config.js
@@ -61,6 +61,9 @@ const nextConfig = {
     return {
       beforeFiles: [
         {
+          // In the future, this rewrite can be generalized by doing something like
+          // source: '/annual-reports/quansight-labs-annual-report-:year.pdf'
+          // destination: '/api/annual-report.pdf?year=:year'
           source: '/annual-reports/quansight-labs-annual-report-2022.pdf',
           destination: '/api/annual-report.pdf',
         },
@@ -77,7 +80,7 @@ const nextConfig = {
           // you choose something like analytics or plausible, it might get
           // blocked in the future.
           source: '/p7e/js/script.js',
-          destination: 'https://plausible.io/js/script.file-downloads.js',
+          destination: 'https://plausible.io/js/script.js',
         },
         {
           source: '/p7e/api/event',
--- a/apps/labs/next.config.js
+++ b/apps/labs/next.config.js
@@ -59,6 +59,12 @@ const nextConfig = {
     // The conventions of this return value are described in the Next.js docs:
     // https://nextjs.org/docs/api-reference/next.config.js/rewrites
     return {
+      beforeFiles: [
+        {
+          source: '/annual-report.pdf',
+          destination: '/api/annual-report.pdf',
+        },
+      ],
       afterFiles: [
         // These rewrites are checked after pages/public files are checked but
         // before dynamic routes.
--- a/apps/labs/next.config.js
+++ b/apps/labs/next.config.js
@@ -71,7 +71,7 @@ const nextConfig = {
           // you choose something like analytics or plausible, it might get
           // blocked in the future.
           source: '/p7e/js/script.js',
-          destination: 'https://plausible.io/js/script.js',
+          destination: 'https://plausible.io/js/script.file-downloads.js',
         },
         {
           source: '/p7e/api/event',
--- a/apps/labs/next.config.js
+++ b/apps/labs/next.config.js
@@ -61,7 +61,7 @@ const nextConfig = {
     return {
       beforeFiles: [
         {
-          source: '/annual-report.pdf',
+          source: '/annual-reports/quansight-labs-annual-report-2022.pdf',
           destination: '/api/annual-report.pdf',
         },
       ],
--- a/apps/labs/pages/api/annual-report.pdf.ts
+++ b/apps/labs/pages/api/annual-report.pdf.ts
@@ -28,6 +28,8 @@ export default async function handler(req: NextRequest) {
       referrer: req.headers.get('referer'),
     }),
   });
+  // Returning fetch comes from Next.js docs:
+  // https://nextjs.org/docs/api-routes/edge-api-routes#forwarding-headers
   return fetch(
     // TODO change this to the real annual report PDF
     'https://a.storyblok.com/f/152463/x/1752e51fa9/nf-annual-report-2021.pdf',
--- a/apps/labs/pages/api/annual-report.pdf.ts
+++ b/apps/labs/pages/api/annual-report.pdf.ts
@@ -22,7 +22,7 @@ export default async function handler(req: NextRequest) {
       'Content-Type': 'application/json',
     },
     body: JSON.stringify({
-      domain: 'gabalafou.com',
+      domain: 'labs.quansight.org',
       name: 'pageview',
       url: req.url,
       referrer: req.headers.get('referer'),
--- a/apps/labs/pages/api/annual-report.pdf.ts
+++ b/apps/labs/pages/api/annual-report.pdf.ts
@@ -25,13 +25,20 @@ export default async function handler(req: NextRequest) {
       domain: 'labs.quansight.org',
       name: 'pageview',
       url: req.url,
+      // A note about spelling: `referrer` on the left (double r) is from the
+      // Plausible spec: https://plausible.io/docs/events-api. `referer` on the
+      // right (single r) is from the HTTP spec. To read more:
+      // https://en.wikipedia.org/wiki/HTTP_referer#Etymology
       referrer: req.headers.get('referer'),
     }),
   });
-  // Returning fetch comes from Next.js docs:
-  // https://nextjs.org/docs/api-routes/edge-api-routes#forwarding-headers
-  return fetch(
-    // TODO change this to the real annual report PDF
+  // A few reasons to use 302 vs 301:
+  // - Ensures that this handler gets called every time that somebody clicks the
+  //   tracking URL for the PDF even if they've clicked it before.
+  // - It signals to search engines and other tech that we do not consider the
+  //   CDN URL for the annual report to be the canonical URL for this resource.
+  return Response.redirect(
     'https://a.storyblok.com/f/152463/x/1752e51fa9/nf-annual-report-2021.pdf',
+    302,
   );
 }
--- a/apps/labs/pages/api/annual-report.pdf.ts
+++ b/apps/labs/pages/api/annual-report.pdf.ts
@@ -12,6 +12,8 @@ export default async function handler(req: NextRequest) {
   req.ip ${req.ip}
   user-agent ${req.headers.get('user-agent')}
   `);
+  // The shape of this fetch comes from the Plausible docs:
+  // https://plausible.io/docs/events-api
   await fetch('https://plausible.io/api/event', {
     method: 'POST',
     headers: {
--- a/apps/labs/pages/api/annual-report.pdf.ts
+++ b/apps/labs/pages/api/annual-report.pdf.ts
@@ -1,0 +1,33 @@
+import type { NextRequest } from 'next/server';
+
+export const config = {
+  runtime: 'experimental-edge',
+};
+
+export default async function handler(req: NextRequest) {
+  // TODO - remove after checking the logs on Vercel
+  console.log(`
+  req.url ${req.url}
+  req.referer ${req.headers.get('referer')}
+  req.ip ${req.ip}
+  user-agent ${req.headers.get('user-agent')}
+  `);
+  await fetch('https://plausible.io/api/event', {
+    method: 'POST',
+    headers: {
+      'User-Agent': req.headers.get('user-agent'),
+      'X-Forwarded-For': req.ip,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      domain: 'gabalafou.com',
+      name: 'pageview',
+      url: req.url,
+      referrer: req.headers.get('referer'),
+    }),
+  });
+  return fetch(
+    // TODO change this to the real annual report PDF
+    'https://a.storyblok.com/f/152463/x/1752e51fa9/nf-annual-report-2021.pdf',
+  );
+}


rgommers:
Thanks Gabriel!

> `labs.quansight.org/annual-report.pdf`

This scheme needs a year, or we'll have a problem next year. Can this be something like `labs.quansight.org/annual-reports/annual-report-2022.pdf`, or `labs.quansight.org/annual-report-2022.pdf`? Once we have multiple reports, we will want an overview page I think, so the former may be preferred.

gabalafou:
@rgommers, done! I used the first pattern you suggested; however, I changed it slightly:

- labs.quansight.org/annual-reports/quansight-labs-annual-report-2022.pdf

That way, if the user decides to download the PDF on their machine and is lazy like me and doesn't want to give the file a different name than the one that the site provides, the file will be saved as `quansight-labs-annual-report-2022.pdf` in their downloads folder. If that doesn't work, it's a very easy, quick change for me to use the route that you first provided verbatim, just lemme know 😊

rgommers:
Even better, thanks @gabalafou!

Katebrack:
@gabalafou this all sounds good on my end and I can see why you need a test email/ad. I will work on getting you an email (I'm guessing a draft will work?) tomorrow with a utm code. 

re: not having data on device-type is fine. I'm curious what it'll come up as in Plausible...perhaps "unknown" but regardless, we'll be fine without that for now.

No other immediate questions, but I'll ping you if any come up. Appreciate the solution you came up with!

gabalafou:
Re: device type, normally what you see in Plausible is a bar chart of the number of visits of each device type, which the following screenshot shows:



But when you filter the data on the PDF url, you get a single, unlabeled, full-width bar:

apps/labs/pages/api/annual-report.pdf.ts
```diff
@@ -0,0 +1,33 @@
...
+  await fetch('https://plausible.io/api/event', {
+    method: 'POST',
+    headers: {
+      'User-Agent': req.headers.get('user-agent'),
+      'X-Forwarded-For': req.ip,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      domain: 'gabalafou.com',
+      name: 'pageview',
```
gabalafou, line 24: I wasn't exactly sure which Plausible event to use for this. I based my decision from reading the [docs page on the Events API](https://plausible.io/docs/events-api). My take was that if I use the `pageview` event, then it will be slotted into the Plausible UI just like other pageviews. [Based on my testing](https://plausible.io/labs.quansight.org?entry_page=%2Fapi%2Fannual-report.pdf), that seems to be the case.

apps/labs/pages/api/annual-report.pdf.ts
```diff
@@ -0,0 +1,33 @@
...
+export const config = {
+  runtime: 'experimental-edge',
+};
+
+export default async function handler(req: NextRequest) {
+  // TODO - remove after checking the logs on Vercel
+  console.log(`
+  req.url ${req.url}
+  req.referer ${req.headers.get('referer')}
+  req.ip ${req.ip}
```
gabalafou, line 12: This is the value I am most curious to check in the Vercel logs because it was turning up empty for me when testing locally (which I think is to be expected but I'm not 100% sure).

apps/labs/pages/api/annual-report.pdf.ts
```diff
@@ -0,0 +1,33 @@
...
+  user-agent ${req.headers.get('user-agent')}
+  `);
+  await fetch('https://plausible.io/api/event', {
```
gabalafou, line 15:
```suggestion
  // The shape of this fetch comes from the Plausible docs:
  // https://plausible.io/docs/events-api
  await fetch('https://plausible.io/api/event', {
```

apps/labs/pages/api/annual-report.pdf.ts
```diff
@@ -0,0 +1,35 @@
...
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      domain: 'gabalafou.com',
+      name: 'pageview',
+      url: req.url,
+      referrer: req.headers.get('referer'),
+    }),
+  });
+  return fetch(
```
gabalafou, line 31:
```suggestion
  // Returning fetch comes from Next.js docs:
  // https://nextjs.org/docs/api-routes/edge-api-routes#forwarding-headers
  return fetch(
```

apps/labs/pages/api/annual-report.pdf.ts
```diff
@@ -0,0 +1,37 @@
...
+    body: JSON.stringify({
+      domain: 'gabalafou.com',
+      name: 'pageview',
+      url: req.url,
+      referrer: req.headers.get('referer'),
+    }),
+  });
+  // Returning fetch comes from Next.js docs:
+  // https://nextjs.org/docs/api-routes/edge-api-routes#forwarding-headers
+  return fetch(
```
gabalafou, line 33: Returning the PDF (rather than returning a redirect to the CDN URL to the PDF) in this Edge API Route might be the wrong way to do this. I just preferred the end user experience. You give the user a URL to the PDF, they go to it, and it loads under that URL, rather than redirecting and loading under the CDN URL. One effect of my decision is that if the user goes to copy the link, the link in their browser remains the same. (Which may or may not be desired from a marketing perspective.) 

For example, if I give an end user the following URL:

- labs.quansight.org/annual-reports/quansight-labs-annual-report-2022.pdf?utm_source=EthicalAds&utm_campain=AnnualReport

Then if the user copies the URL from their browser address bar and pastes it to a friend, it will be exactly the same URL.

My understanding is that doing it this way will increase our billable Vercel usage. That's because (if I understand correctly) Vercel's servers will be fetching and returning the PDF data through the URL route. But my understanding is that this route is temporary as we will eventually have an HTML page that contains the report, in which we can link directly to the CDN URL for the PDF and eventually delete this route altogether.

gabalafou (commented):
Okay done self-reviewing

trallard:
Thanks for the detailed explanation @gabalafou 

The updated URL with the year and the improved PDF name seems excellent to me and would future-proof us as we plan to make this a yearly thing.

> As far as I can tell, we get all of the same data using this method that we would get with an HTML page EXCEPT for what kind of device the user was using (mobile vs tablet vs desktop).

This does not seem like a big problem (at least to me) - and if this means we avoid accessibility anti-patterns, I am +1 on keeping this approach. 
Additionally, giving the reader a choice between downloading or not a file (or redirecting to the PDF) is preferable and avoids surprises on the reader's side.

apps/labs/pages/api/annual-report.pdf.ts
trallard: With the updated [URL this seems to be the new endpoint](https://plausible.io/labs.quansight.org?entry_page=%2Fapi%2Fquansight-labs-annual-report-2022.pdf)

This seems like a reasonable event

apps/labs/pages/api/annual-report.pdf.ts
```diff
@@ -0,0 +1,37 @@
...
+  // https://plausible.io/docs/events-api
+  await fetch('https://plausible.io/api/event', {
+    method: 'POST',
+    headers: {
+      'User-Agent': req.headers.get('user-agent'),
+      'X-Forwarded-For': req.ip,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      domain: 'gabalafou.com',
```
trallard, line 25: Is this meant to be a placeholder domain? I assume this will need to be swapped to the labs domain @gabalafou

apps/labs/pages/api/annual-report.pdf.ts
trallard: Interesting 🤔 - @Katebrack is the `HTML` page referred to above a page where we will be linking to the various years' reports as in the NumFOCUS site?

apps/labs/pages/api/annual-report.pdf.ts
gabalafou: Good catch! Yes. This was me trying not to pollute our labs.quansight.org, but eventually I gave up because Plausible doesn't really support a dev-only account. You gotta pay. 
```suggestion
      domain: 'labs.quansight.org',
```

apps/labs/pages/api/annual-report.pdf.ts
gabalafou: Slight correction. After Ralf's requested change, I think the Plausible dashboard URL should look like this:

https://plausible.io/labs.quansight.org?entry_page=%2Fannual-reports%2Fquansight-labs-annual-report-2022.pdf

apps/labs/pages/api/annual-report.pdf.ts
Katebrack: To be clear, we're not planning to create an HTML version and build out a page similar to NumFOCUS until ~next year. All we need for this year is the PDF link which, from what I read, we have

apps/labs/next.config.js
```diff
@@ -71,7 +77,7 @@ const nextConfig = {
           // you choose something like analytics or plausible, it might get
           // blocked in the future.
           source: '/p7e/js/script.js',
-          destination: 'https://plausible.io/js/script.js',
+          destination: 'https://plausible.io/js/script.file-downloads.js',
```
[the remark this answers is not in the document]
bskinn: > I can't think of a reason why we would need to link directly to the CDN file

Nor can I. With this machinery wiring the downloads into the pageviews metrics, I don't think we need to add this slice of the Plausible tooling. OK to revert.

apps/labs/pages/api/annual-report.pdf.ts
bskinn: > My understanding is that doing it this way will increase our billable Vercel usage.

How close are we coming to a breakpoint on our Vercel usage at present? If we're well within our current allocation (if we have one?), I don't think this is an issue.

On the other hand, if we *are* paying a pro rata on bandwidth usage, then we should consider resizing some images: https://vercel.com/dashboard/quansight/usage?topPathsFullscreen=true&pathType=source_path

apps/labs/pages/api/annual-report.pdf.ts
```diff
@@ -0,0 +1,37 @@
...
+    headers: {
+      'User-Agent': req.headers.get('user-agent'),
+      'X-Forwarded-For': req.ip,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      domain: 'labs.quansight.org',
+      name: 'pageview',
+      url: req.url,
+      referrer: req.headers.get('referer'),
```
bskinn, line 28: This is spelled `referer` everywhere in the file except here.

Are we API-constrained to use `referrer` as the key here, or could we harmonize to `referer`?

If we can't harmonize, I'd like a comment here to call out the intentional difference.

apps/labs/pages/api/annual-report.pdf.ts
bskinn: I prefer the UX this way, too.

bskinn (commented):
Couple of comments/questions, but overall makes sense and LGTM.

Did not dig into the APIs for correctness. Figure it's been easy enough to test the download and the analytics ping manually.

apps/labs/pages/api/annual-report.pdf.ts
gabalafou: I changed it to use a redirect because that seems safer long-term.

bskinn: Causing the user's browser to change `.location` to the asset, instead of retrieving the asset server-side...which means that the users location bar will show the storyblok.com URL.

This clearly works, so I'm good with it functionally. What about the redirect makes it safer than the direct fetch?

bskinn (approved):
Content with the technical solution. [Raised the question](https://github.com/Quansight/Quansight-website/pull/645#discussion_r1096240697) of 'a redirect is safer than a direct fetch`, but I don't need to hold up the PR for that answer.

apps/labs/pages/api/annual-report.pdf.ts
gabalafou: Safer in the face of ignorance. Vercel puts usage limits (timeouts, bandwidth) on these Edge API Routes functions. I assume that pumping the PDF through the handler would be fine, but this errs on the side of "I don't want to have to go fix it" if my assumption is somehow wrong.

bskinn: :+1:, the redirect makes it "more normal" traffic, and less likely to make Vercel cranky. Got it.

gabalafou: Yeah, it's like I'm 99.9% sure the redirect is a safe and sound way to do this, versus I'm like 90% sure about the other way. And I've already put so much time into this, and there are still some other unverified assumptions that need to be checked, so this is me trying to not create anymore work for myself as well as giving myself some piece of mind.

bskinn: Yeah, I could see the fetch working great during our testing... and then once it goes live and a few hundred people (hopefully!) hit it over a few days Vercel then going "Nopers!" and kablooey.

PR #647, opened 2023-02-01
[BLOG] Add uarray-intro

MarsBarLee:
Link to [blog post, successful build](https://labs-git-blog-name-quansight.vercel.app/blog/blog-name).

## Text styling

## Non-text contents


apps/labs/posts/uarray-intro.md
```diff
@@ -0,0 +1,331 @@
...
+# But works if we do this:
+
+with ua.set_backend(be1, coerce=True):
+    print(type(myfunc(1.0)))
+
+with ua.set_backend(be2, coerce=True):
+    print(type(myfunc(1.0)))
+```
+
+    <class 'int'>
```
MarsBarLee, line 284: @gabalafou I'm getting the Vercel build error: `error [Error: [next-mdx-remote] error compiling MDX:`

Full error message
```
Unexpected character `'` (U+0027) after name,
expected a character that can start an attribute name,
such as a letter, `$`, or `_`; whitespace before attributes;
or the end of the tag

  270 |
> 271 |     <class 'int'>
      |            ^
  272 |     <class 'str'>
  273 |
```

[Link to error](https://vercel.com/quansight/labs/7bDxZz8gpDGNVX5y1VZmSHTeUNBd#:~:text=15%3A02%3A27.909-,error%20%5BError%3A%20%5Bnext%2Dmdx%2Dremote%5D%20error%20compiling%20MDX%3A,-15%3A02%3A27.910)

The context is showing the result of some code. [Link to this section in the old blog website.](https://labs.quansight.org/blog/2019/04/uarray-intro#:~:text=%3C-,class,-%27int%27%3E%0A%3Cclass%20%27str)

Maybe we could wrap it in a code block? But then it would look out of place with the rest of the code results, unless the rest were also in a code block.

gabalafou: Ugh that is formatted so ugly on the old site. If I understand correctly, and as I think you're saying, on this blog post on the old website, the blocks with gray background are meant to represent the code blocks, whereas the monospace-font lines (with no background color) that immediately follow are meant to represent the output from running those code blocks. But the way that is shown visually in the old blog post isn't intuitive to me at all. Let me play with this a bit and get back to you. Thanks for providing the link to the old blog post, by the way, super helpful!

MarsBarLee: Thanks for taking a look at it! And yes, I understood the blog the same way, code and then result. Maybe both could be in code blocks, but have different color backgrounds or text, to emphasize the difference?

PR #648, opened 2023-02-01
Release 2023-02-01

gabalafou:
Bring the changes from `develop` into `main`

PR #650, opened 2023-02-03
Fix path matcher

gabalafou:
Before: match any route that ends in py or cpp
After: match any route that ends in .py or .cpp

This fixes an issue where /blog/re-engineering-cicd-pipelines-for-scipy was returning as plain text rather than as an HTML web page. See:

- https://github.com/Quansight/Quansight-website/pull/648#issuecomment-1412956951

Here's how to test. Visit the following URLs. The first should load as a regular web page. The second and third should load as plain text file (not trigger a download file dialog).

1. https://labs-git-fix-scipy-pipelines-quansight.vercel.app/blog/re-engineering-cicd-pipelines-for-scipy
2. https://labs-git-fix-scipy-pipelines-quansight.vercel.app/posts/cxx-numba-interoperability/cxx2py.py
3. https://labs-git-fix-scipy-pipelines-quansight.vercel.app/posts/cxx-numba-interoperability/cxx2py_libfoo.cpp

### DIFF
--- a/apps/labs/next.config.js
+++ b/apps/labs/next.config.js
@@ -35,8 +35,10 @@ const nextConfig = {
         //   Content-Type: application/octet-stream
         //
         // The following rule catches some plain text file types that we link
-        // to, and supplies the text/plain MIME type in the Content-Type header.
-        source: '/:path*(.py|.cpp)',
+        // to, and supplies the text/plain MIME type in the Content-Type header
+        // so that end users can open and view the file directly in their
+        // browser rather than having to first download and save the file.
+        source: '/:path*.(py|cpp)',
         headers: [
           {
             key: 'Content-Type',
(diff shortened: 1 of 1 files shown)


bskinn:
Confirm that the three test links work as advertised. :+1:

apps/labs/next.config.js
```diff
@@ -35,8 +35,10 @@ const nextConfig = {
         //   Content-Type: application/octet-stream
         //
         // The following rule catches some plain text file types that we link
-        // to, and supplies the text/plain MIME type in the Content-Type header.
-        source: '/:path*(.py|.cpp)',
+        // to, and supplies the text/plain MIME type in the Content-Type header
+        // so that end users can open and view the file directly in their
+        // browser rather than having to first download and save the file.
+        source: '/:path*.(py|cpp)',
```
bskinn, line 41: May want to add a comment here as to why this pattern works... that only the stuff inside the parentheses is interpreted as regex, and thus the period outside the parens is a period literal, not a regex wildcard.

Could also link to the [relevant section of Next docs](https://nextjs.org/docs/api-reference/next.config.js/headers#header-overriding-behavior) for good measure?

bskinn (approved):
Functionality looks good. Could use a bit of extra commentary, see review note in code.

PR #662, opened 2023-02-09
Fix test-homepage by fixing paths filter

gabalafou:
@bskinn discovered that /test-homepage in the Consulting Storyblok was not getting deployed to a Vercel preview build. 

I think it's because of the way paths were being filtered in the app, rejecting any slug if it it contained any word from a restricted list. One of the words on that list is 'homepage,' so it would match 'test-homepage' (unless I've misread the code somehow).

### DIFF
--- a/apps/consulting/utils/getArticlesPaths/getArticlesPaths.ts
+++ b/apps/consulting/utils/getArticlesPaths/getArticlesPaths.ts
@@ -2,14 +2,14 @@ import { TLinkEntry } from '@quansight/shared/types';
 
 import { LinkEntry } from '../../api/types/basic';
 import { formatSlugParam } from './formatSlugParam';
-import { isSlugRestricted } from './isSlugRestricted';
+import { isSlugForPost } from './isSlugForPost';
 import { TGetArticlesPaths } from './types';
 
 export const getArticlesPaths = (
   items: TLinkEntry<LinkEntry>[],
 ): TGetArticlesPaths[] =>
   items
-    .filter(({ isFolder, slug }) => slug && !isFolder && isSlugRestricted(slug))
+    .filter(({ isFolder, slug }) => slug && !isFolder && isSlugForPost(slug))
     .map(({ slug }) => ({
       params: {
         slug: formatSlugParam(slug),
--- a/apps/consulting/utils/getArticlesPaths/isSlugForPost.ts
+++ b/apps/consulting/utils/getArticlesPaths/isSlugForPost.ts
@@ -1,0 +1,4 @@
+import { ARTICLES_DIRECTORY_SLUG } from './constants';
+
+export const isSlugForPost = (slug: string): boolean =>
+  slug.startsWith(ARTICLES_DIRECTORY_SLUG);
--- a/libs/shared/utils/src/getPaths/getPaths.ts
+++ b/libs/shared/utils/src/getPaths/getPaths.ts
@@ -1,15 +1,12 @@
 import { isSlugRestricted } from './isSlugRestricted';
 import { TGetPaths } from './types';
 
-export const getPaths = <
-  LinkEntry extends { isFolder: boolean; slug: string; parentId: number },
->(
+export const getPaths = <LinkEntry extends { isFolder: boolean; slug: string }>(
   items: LinkEntry[],
 ): TGetPaths[] =>
   items
     .filter(
-      ({ isFolder, slug, parentId }) =>
-        slug && !isFolder && !parentId && !isSlugRestricted(slug),
+      ({ isFolder, slug }) => slug && !isFolder && !isSlugRestricted(slug),
     )
     .map(({ slug }) => ({
       params: {
--- a/libs/shared/utils/src/getPaths/getPaths.ts
+++ b/libs/shared/utils/src/getPaths/getPaths.ts
@@ -1,12 +1,15 @@
 import { isSlugRestricted } from './isSlugRestricted';
 import { TGetPaths } from './types';
 
-export const getPaths = <LinkEntry extends { isFolder: boolean; slug: string }>(
+export const getPaths = <
+  LinkEntry extends { isFolder: boolean; slug: string; parentId: number },
+>(
   items: LinkEntry[],
 ): TGetPaths[] =>
   items
     .filter(
-      ({ isFolder, slug }) => slug && !isFolder && !isSlugRestricted(slug),
+      ({ isFolder, slug, parentId }) =>
+        slug && !isFolder && !parentId && !isSlugRestricted(slug),
     )
     .map(({ slug }) => ({
       params: {
--- a/libs/shared/utils/src/getPaths/getPaths.ts
+++ b/libs/shared/utils/src/getPaths/getPaths.ts
@@ -5,7 +5,9 @@ export const getPaths = <LinkEntry extends { isFolder: boolean; slug: string }>(
   items: LinkEntry[],
 ): TGetPaths[] =>
   items
-    .filter(({ isFolder, slug }) => slug && !isFolder && isSlugRestricted(slug))
+    .filter(
+      ({ isFolder, slug }) => slug && !isFolder && !isSlugRestricted(slug),
+    )
     .map(({ slug }) => ({
       params: {
         slug,
--- a/libs/shared/utils/src/getPaths/getPaths.ts
+++ b/libs/shared/utils/src/getPaths/getPaths.ts
@@ -6,7 +6,8 @@ export const getPaths = <LinkEntry extends { isFolder: boolean; slug: string }>(
 ): TGetPaths[] =>
   items
     .filter(
-      ({ isFolder, slug }) => slug && !isFolder && !isSlugRestricted(slug),
+      ({ isFolder, slug }) =>
+        slug && !isFolder && !slug.includes('/') && !isSlugRestricted(slug),
     )
     .map(({ slug }) => ({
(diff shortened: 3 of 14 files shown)


gabalafou:
I haven't tested this PR and we will need to test it rather thoroughly before merging because it could really break things.

apps/consulting/utils/getArticlesPaths/isSlugForPost.ts
```diff
@@ -0,0 +1,4 @@
+import { ARTICLES_DIRECTORY_SLUG } from './constants';
+
+export const isSlugForPost = (slug: string): boolean =>
+  slug.startsWith(ARTICLES_DIRECTORY_SLUG);
```
bskinn, line 4: Interesting. I kind of would think that this would fix this:

image

But I guess not.

libs/shared/utils/src/getPaths/isSlugRestricted.ts
```diff
@@ -1,7 +1,7 @@
 import { restrictedSlugs } from './restrictedSlugs';
```
bskinn, line 1: Maybe this is where the fix needs to happen?

'post' doesn't appear in `restrictedSlugs` -- it [instead (still?) has `blog`](https://github.com/Quansight/Quansight-website/blob/86078fe8516aa9e3dfa863a5479cc000c475712d/libs/shared/utils/src/getPaths/restrictedSlugs.ts#L10), which I think was relevant before we located the blog posts to `/post`, and is out of date now.

bskinn (commented):
Some thoughts related to the issue of the blog posts getting double-built, once at `post/{postname}` and once at `post%2F{postname}`.

apps/consulting/utils/getArticlesPaths/getArticlesPaths.ts
```diff
@@ -2,14 +2,14 @@ import { TLinkEntry } from '@quansight/shared/types';
 
 import { LinkEntry } from '../../api/types/basic';
 import { formatSlugParam } from './formatSlugParam';
-import { isSlugRestricted } from './isSlugRestricted';
+import { isSlugForPost } from './isSlugForPost';
```
bskinn, line 5: It looks like this is a cosmetic+semantic name change, to differentiate this from the `isSlugRestricted` elsewhere in the PR?

libs/shared/utils/src/getPaths/isSlugRestricted.ts
```diff
@@ -1,7 +1,7 @@
...
 
 export const isSlugRestricted = (slug: string): boolean => {
-  return !restrictedSlugs.some((restrictedSlug) =>
-    slug.includes(restrictedSlug),
+  return restrictedSlugs.some((restrictedSlug) =>
+    new RegExp(restrictedSlug, 'i').test(slug),
```
bskinn, line 5: Is there any particular functional advantage to the switch to a regex here?

For now, the `restrictedSlugs` are all plain strings. I'm guessing this is in anticipation of wanting to tighten the matching on `restrictedSlugs` in the future (or even the present)?

bskinn (commented):
Couple of questions

libs/shared/utils/src/getPaths/isSlugRestricted.ts
bskinn: Oooh, now I see the odd semantics that were here before... I agree, it makes much more sense for the `.some()` ***not*** to be negated within `isSlugRestricted()` here but instead to have the negation at the point of use in `getPaths.ts`.

apps/consulting/utils/getArticlesPaths/getArticlesPaths.ts
gabalafou: It's not to differentiate. Given what the function does, the name simply made no sense to me, so I decided to change it.

apps/consulting/utils/getArticlesPaths/isSlugForPost.ts
gabalafou: It doesn't fix that issue because this function isn't used in `getPaths()`. I started to slip that fix into this PR but then it felt like I was putting too much into one PR. I will add a comment below showing you in the code how this can be fixed.

libs/shared/utils/src/getPaths/getPaths.ts
```diff
@@ -5,7 +5,9 @@ export const getPaths = <LinkEntry extends { isFolder: boolean; slug: string }>(
   items: LinkEntry[],
 ): TGetPaths[] =>
   items
-    .filter(({ isFolder, slug }) => slug && !isFolder && isSlugRestricted(slug))
+    .filter(
+      ({ isFolder, slug }) => slug && !isFolder && !isSlugRestricted(slug),
```
gabalafou, line 9: Here's how you fix the other issue:

```suggestion
      ({ isFolder, slug }) => 
        slug && !isFolder && !isSlugForPost(slug) && !isSlugRestricted(slug),
```

libs/shared/utils/src/getPaths/isSlugRestricted.ts
gabalafou: You're right. We don't need the RegExp for case insensitivity. I will change it.

gabalafou, line 5:
```suggestion
    slug.toLowerCase() === restrictedSlug,
```

gabalafou: Also what I wrote is just plain wrong, i.e., won't fix the test-page issue. 😅

libs/shared/utils/src/getPaths/isSlugRestricted.ts
gabalafou: - https://github.com/Quansight/Quansight-website/pull/663/files

apps/consulting/utils/getArticlesPaths/getArticlesPaths.ts
gabalafou: Differentiation is just a side benefit 😃

libs/shared/utils/src/getPaths/isSlugRestricted.ts
bskinn: :+1: to removing `'blog'` ... Should we also add `'^/post/`?

That may fix the post double-rendering.

Or, it may entirely break the blog build?

bskinn: Ahhhh. Nevermind, I made it to the end of the PR thread. Got it.

libs/shared/utils/src/getPaths/isSlugRestricted.ts
```diff
@@ -1,7 +1,7 @@
 import { restrictedSlugs } from './restrictedSlugs';
 
 export const isSlugRestricted = (slug: string): boolean => {
-  return !restrictedSlugs.some((restrictedSlug) =>
-    slug.includes(restrictedSlug),
+  return restrictedSlugs.some(
+    (restrictedSlug) => slug.toLowerCase() === restrictedSlug,
```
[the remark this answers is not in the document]
bskinn: Ah. Regex for the win, then. They should probably all be e.g. `^/home`, `^/team-members/`, etc., I figure.

gabalafou (commented):
Okay, I've simplified things and solved multiple problems. If Vercel preview builds don't fail, then this is ready for review again.

gabalafou:
Here's a [dump of all the links](https://gist.github.com/gabalafou/756bf482971b68d0751889c6cc4f7a2b)

libs/shared/utils/src/getPaths/getPaths.ts
```diff
@@ -5,7 +5,10 @@ export const getPaths = <LinkEntry extends { isFolder: boolean; slug: string }>(
   items: LinkEntry[],
 ): TGetPaths[] =>
   items
-    .filter(({ isFolder, slug }) => slug && !isFolder && isSlugRestricted(slug))
+    .filter(
+      ({ isFolder, slug }) =>
+        slug && !isFolder && !slug.includes('/') && !isSlugRestricted(slug),
```
bskinn, line 10: This makes sense, at least for now. I don't anticipate us wanting a path hierarchy for regular site pages... easy enough to change if we *do* end up wanting it, though.

libs/shared/utils/src/getPaths/restrictedSlugs.ts
```diff
@@ -1,10 +1,12 @@
...
-  'team',
-  'layout',
+  'about-us',
   'library',
-  'team-members',
-  'about',
+  // Labs
   'home',
   'team',
   'blog',
```
bskinn, line 12: Ooooh, I didn't even think about this being both sites. Good call, esp adding the comments.

libs/shared/utils/src/getPaths/restrictedSlugs.ts
```diff
@@ -1,10 +1,12 @@
+// These are all of the Storyblok slugs that should not match pages/[slug]
+// because they are matched to other pages. For example, Labs 'blog' is matched
+// to apps/labs/pages/blog/index.tsx.
```
bskinn, line 3:
```suggestion
// to apps/labs/pages/blog/index.tsx.
// Note that all these names will be excluded from pages/[slug]
// on *both* sites, since this is shared code.
```

gabalafou:
old, stale

PR #674, opened 2023-02-22
[DESIGN] Change Features component styling to better match new "Our Values" icons

bskinn:
Changes include:

- Switch background color to violet of brand theme: `#452393`.

Attn: @Katebrack

### DIFF
--- a/apps/consulting/components/Features/Features.tsx
+++ b/apps/consulting/components/Features/Features.tsx
@@ -5,7 +5,7 @@ import { TFeaturesProps } from './types';
 
 export const Features: FC<TFeaturesProps> = ({ title, columns }) => {
   return (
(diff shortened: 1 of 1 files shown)


bskinn:
@pavithraes or @gabalafou Can you review and merge this so that it goes live with the next build?

gabalafou (changes requested):
Oops. Sorry I approved this too quickly after looking at the current website about us and the preview about us. 

But I need to do quick check on where else this component is used.

gabalafou (approved):
Filtering Storyblok content by the Features block shows that the only published page in the Consulting space that uses the Features block is the About Us page, as the following screenshot shows:



So this is good to go.

bskinn:
> Filtering Storyblok content by the Features block

Nice, I wondered if this was possible. I did a manual page-by-page check to confirm this was the only point of use.

If we need to use it multiple places later on, adding a color toggle might be a good FEAT. But not now.

PR #712, opened 2023-04-05
[CONS BLOG] Add 'Scalable Computing' category to 'Evolving QHub to Nebari' post

bskinn:
The 'Announcing QHub' is in the 'scalable computing' category; it seems that the Nebari post should be also.


bskinn:
@pavithraes What do you think about adding the 'Scalable Computing' tag to the 'Evolving QHub to Nebari' post?

If you think it fits, we might as well include this in today's build to live.

pavithraes (approved):
Good idea, thank you!

PR #714, opened 2023-04-05
Change preview header so it doesn't mess up layout

gabalafou:


gabalafou:
Best way to test this PR is to look at the preview builds on desktop and mobile and see how the preview banner looks and feels with these changes

bskinn (approved):
I encountered this while reviewing the Nebari Services page
and during work on the header nav.

Functionally, seems good; aesthetically, seems good.

Didn't closely review the code, but given that it's functioning
reasonably I didn't see a need to.

PR #719, opened 2023-04-12
On Nebari Services page, more hit area on call-to-action buttons

gabalafou:
I noticed an issue with my first implementation of the "Contact Sales" and other buttons on the Nebari Services page. The whole button area should be clickable but it currently isn't. This PR fixes that.


bskinn:
PR also changes how the right-pointing triangle in the CTA text is defined -- previously it was inserted using a Tailwind `after:content-` class; now it's an explicit element in the page source. 

Much easier to tell where the triangle comes from this way. :+1:

bskinn (approved):
Multiple replicas of the same code change, putting the Tailwind styling inside the `<Link>`, instead of outside... causing the full color field of the button to be clickable.

All buttons behave as expected and all looks ok.

LGTM.

PR #728, opened 2023-04-26
Fix form prepopulate bug

gabalafou:
The form library we use makes it difficult (impossible?) to programmatically change the value of a form field from outside the component and have the form pass validation.

So this PR moves that field prepopulation logic into the Form component and uses session storage as the place to get a prepopulate value from the outside.

### DIFF
--- a/libs/shared/ui-components/src/Form/Form.tsx
+++ b/libs/shared/ui-components/src/Form/Form.tsx
@@ -1,4 +1,4 @@
-import { FC, useState } from 'react';
+import { FC, useState, useEffect } from 'react';
 
 import clsx from 'clsx';
 import { useForm } from 'react-hook-form';
@@ -18,7 +18,7 @@ import { FormError } from './FormError';
 import { FormHeader } from './FormHeader';
 import { FormImage } from './FormImage';
 import { FormSuccess } from './FormSuccess';
-import { FormStates, TFormProps } from './types';
+import { FormStates, TFormProps, FormSessionStorageKeys } from './types';
 import { getFormHeader, getTrackingParams } from './utils';
 
 export const backgroundStyles = `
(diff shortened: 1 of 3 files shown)


libs/shared/ui-components/src/Form/Form.tsx
```diff
@@ -32,8 +32,20 @@ export const Form: FC<TFormProps> = (props) => {
   const {
     register,
     handleSubmit,
+    setValue,
     formState: { errors, isValid },
-  } = useForm<FormValues>({ mode: 'onChange' });
+  } = useForm<FormValues>({ mode: 'onSubmit' });
```
bskinn, line 37: If I read the docs right, this change should make it so that it only validates form contents on submit attempts?

If so, something is weird, though not necessarily a merge blocker. After I attempt to submit the form with 1+ fields empty, I get the red "This field is required" message under the empty ones. Then, if I add content to one of them, the red "This field is required" goes away immediately, without my clicking 'Submit'. If I re-delete that new content, the red "This field is required" returns immediately, without my clicking 'Submit'.

Is it perhaps the case that `mode: 'onChange'` behavior is enabled after a failed form submit?

bskinn (approved):
I don't fully understand the JS code here, but it makes sense to within "Ok, this is probably a React idiom for such-and-such" and I *am* able to submit the form with an unmodified canned-populated message from the Nebari support page.

Added a couple of single questions; but, none are blockers to merge.

libs/shared/ui-components/src/Form/Form.tsx
gabalafou: Yeah, I think the part of the react-hook-form API that explains what you're seeing is an option called `reValidateMode` (default is `onChange`).

Also, this change slipped through the cracks. My first attempt at fixing the bug was to change the mode from `onChange` to `onSubmit`, but that didn't do the trick, and I realized that there was probably no easy way to get the Form component to re-validate the form after I changed a value via `messageField.value = msg`. 

I can revert this change if you prefer.

gabalafou:
Good review comments! I changed the code, tested it on the preview builds, so I'm merging it now

PR #733, opened 2023-05-26
[FEAT] Add social card tags

Surafel-Getachew:
Added social card for twitter and other sites that follow open graph protocol


libs/shared/ui-components/src/SEO/SEO.tsx
```diff
@@ -1,14 +1,60 @@
...
+      {/* open-graph / LinkedIn Specification */}
+      <meta property="og:title" content={title} />
+      <meta property="og:description" content={description} />
+      <meta property="og:url" content={url} />
+      <meta
+        property="og:image"
+        content={
+          DomainVariant[variant] === 'Quansight'
+            ? 'https://a.storyblok.com/f/147759/1200x630/ce491624ae/quansight-logo-opengraph.png'
+            : 'https://a.storyblok.com/f/152463/900x472/fe2083b744/qslabs-logo-opengraph.png'
```
bskinn, line 55:
```suggestion
            : 'https://a.storyblok.com/f/152463/900x472/b289c484e3/qslabs-logo-opengraph-size.png'
```

libs/shared/ui-components/src/SEO/SEO.tsx
```diff
@@ -1,14 +1,60 @@
...
+
+      {/* open-graph / LinkedIn Specification */}
+      <meta property="og:title" content={title} />
+      <meta property="og:description" content={description} />
+      <meta property="og:url" content={url} />
+      <meta
+        property="og:image"
+        content={
+          DomainVariant[variant] === 'Quansight'
+            ? 'https://a.storyblok.com/f/147759/1200x630/ce491624ae/quansight-logo-opengraph.png'
```
bskinn, line 54:
```suggestion
            ? 'https://a.storyblok.com/f/147759/1200x630/8b20f86b2a/quansight-logo-opengraph-size.png'
```

libs/shared/ui-components/src/SEO/SEO.tsx
```diff
@@ -1,14 +1,60 @@
...
+
+      {/* open-graph / LinkedIn Specification */}
+      <meta property="og:title" content={title} />
+      <meta property="og:description" content={description} />
+      <meta property="og:url" content={url} />
+      <meta
+        property="og:image"
+        content={
+          DomainVariant[variant] === 'Quansight'
+            ? 'https://a.storyblok.com/f/147759/1200x630/8b20f86b2a/quansight-logo-opengraph-size`.png'
```
bskinn, line 54:
```suggestion
            ? 'https://a.storyblok.com/f/147759/1200x630/8b20f86b2a/quansight-logo-opengraph-size.png'
```

Surafel-Getachew: Oh, I'll update it now thanks

bskinn:
@Surafel-Getachew Not all pages are working correctly. E.g., the homepage (https://consulting-git-fork-surafel-getachew-social-cards-quansight.vercel.app/) parses ok in https://www.linkedin.com/post-inspector/, but https://consulting-git-fork-surafel-getachew-social-cards-quansight.vercel.app/careers does **not** parse ok.

bskinn:
[Twitter social card validator](https://cards-dev.twitter.com/validator)

bskinn:
@Surafel-Getachew Based upon the `quansight.com` domain showing up in an odd place in the 'Canonical URL', I'm guessing that [prepending `origin` to `url`][prepend] isn't the correct approach here?

image


[prepend]: https://github.com/Quansight/Quansight-website/pull/733/files#diff-eb588762b56d32a84842239bf30b2359c73abb1c99c6e69dcbc5ea0e3a193a9aR16

Surafel-Getachew:
> @Surafel-Getachew Based upon the `quansight.com` domain showing up in an odd place in the 'Canonical URL', I'm guessing that [prepending `origin` to `url`](https://github.com/Quansight/Quansight-website/pull/733/files#diff-eb588762b56d32a84842239bf30b2359c73abb1c99c6e69dcbc5ea0e3a193a9aR16) isn't the correct approach here?
> 
> image

I think so @bskinn I think we have to get the origin from the browser.

bskinn:
It looks to me like LinkedIn/OG is expecting a relative URL.

bskinn:
But it's working ok for Labs, for non-root pages... e.g. https://labs-git-fork-surafel-getachew-social-cards-quansight.vercel.app/projects

Surafel-Getachew:
> It looks to me like LinkedIn/OG is expecting a relative URL in `og:url`.

But the recommended approach is to use absolute path

bskinn:
> But the recommended approach is to use absolute path

[Indeed](https://ogp.me/#metadata).

Huh. Is there a different way to get at the full URL of the page? `window.location`, say?

Because, really, I think we'd like this `url` to point to the _actual location_ of each page, whether that's live on `https://quansight.com/careers` or at a preview build, like `https://consulting-git-fork-surafel-getachew-social-cards-quansight.vercel.app/careers`

bskinn:
The page source for Consulting's `.../careers` page has the URL as relative:

image

Perhaps it's necessary to explicitly include the `https://` when composing `og:url`?

Surafel-Getachew:
> The page source for Consulting's `.../careers` page has the URL as relative:
> 
> image
> Perhaps it's necessary to explicitly include the `https://` when composing `og:url`?

that's actually the recommended way, I'll add it

Surafel-Getachew:
> > But the recommended approach is to use absolute path
> 
> [Indeed](https://ogp.me/#metadata).
> 
> Huh. Is there a different way to get at the full URL of the page? `window.location`, say?
> 
> Because, really, I think we'd like this `url` to point to the _actual location_ of each page, whether that's live on `https://quansight.com/careers` or at a preview build, like `https://consulting-git-fork-surafel-getachew-social-cards-quansight.vercel.app/careers`

Yes. But we have to be careful as the window can be undefined especially in next JS apps. So before using the window property I have to verify it works when the app is build. is there a way/script to run the build after building the app?

bskinn:
> But we have to be careful as the window can be undefined especially in next JS apps.

Could we use the live site domains as a fallback? That is, use `window.location` when it's defined, but fall back to `quansight.com` or `labs.quansight.org` (note the `.org` here) when it's not?


> is there a way/script to run the build after building the app?

I'm not sure what you mean by this -- if you mean run the app on a local dev server, there's `npm run start:consulting` and `npm run start:labs`.

If you mean loading the site after a static site build, then `npm run build`, and point a browser at `/dist/apps/{consulting|labs}`.

bskinn:
Looks like `window.location` is indeed not available during the static builds. :-/

image

We might just have to live with all of these pointing at `quansight.com` and `labs.quansight.org`

libs/shared/ui-components/src/SEO/SEO.tsx
```diff
@@ -1,14 +1,63 @@
...
-  <Head>
-    <title>{`${title} | ${DomainVariant[variant]}`}</title>
-    <meta name="description" content={description} />
-  </Head>
-);
+export const SEO: FC<TSEOProps> = ({ title, description, variant }) => {
+  const { asPath } = useRouter();
+  const fallBackOrigin =
+    DomainVariant[variant] === 'Quansight'
+      ? 'quansight.com'
```
bskinn, line 14: I think we need this to be absolute

```suggestion
      ? 'https://quansight.com'
```

libs/shared/ui-components/src/SEO/SEO.tsx
```diff
@@ -1,14 +1,63 @@
...
+      : 'labs.quansight.org';
```
bskinn, line 15: Here too

```suggestion
      : 'https://labs.quansight.org';
```

bskinn (changes requested):
Pretty sure we need the `https://` in these fallbacks

libs/shared/ui-components/src/SEO/SEO.tsx
```diff
@@ -1,14 +1,63 @@
...
 
 import { TSEOProps } from './types';
 
-export const SEO: FC<TSEOProps> = ({ title, description, variant }) => (
-  <Head>
-    <title>{`${title} | ${DomainVariant[variant]}`}</title>
-    <meta name="description" content={description} />
-  </Head>
-);
+export const SEO: FC<TSEOProps> = ({ title, description, variant }) => {
```
bskinn, line 10: @Surafel-Getachew -- While piggybacking this onto the `SEO` component is clearly working, this social preview metadata isn't really part of the SEO of the pages.

Can you re-factor the social media `<meta>` logic into a new component... `SocialMeta`, say?

That would make for a better semantic separation.

Surafel-Getachew: @bskinn, The reason I add it in SEO component is one it's the only component that is shared by all the pages with the information we need for example (title, description, domainVariant).
If we create `SocialMeta` component we have the option
1. To add `SocialMeta` component on every page

bskinn: Then a deeper refactor may be called for... where the component currently called `SEO` gets renamed to something like `HeadMeta`, and the `HeadMeta` includes both `SEO` and `SocialMeta`.

Surafel-Getachew: @bskinn , actually we are going to do that refactor sooner or later because in the future we will not be using Quansight website logo and labs logo for each page. they may have their own images and in that case having a component that accepts the specific image of that page will be needed so it's just a matter of time so I've already implemented it.

Surafel-Getachew:
@bskinn I think https://consulting-iwjkh1hgy-quansight.vercel.app/careers is also working now.

package-lock.json
bskinn, line 1: @Surafel-Getachew -- what happened to `package-lock.json`? It looks like it got inappropriately regenerated or something, and much of its content was lost.

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,10 @@
+import { DomainVariant } from '@quansight/shared/types';
+
+export type TSocailCardProps = {
```
bskinn, line 3: Typo here: `TSocialCardProps` -- will need to be propagated to all imports and points of use

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+        : 'https://a.storyblok.com/f/152463/900x472/b289c484e3/qslabs-logo-opengraph-size.png';
+    }
+  };
+  const renderCardImgAlt = (): string => {
+    if (cardImgAlt) {
+      return cardImgAlt;
+    } else {
+      return DomainVariant[variant] === 'Quansight'
+        ? 'Quansight logo'
+        : 'Quansight labs logo';
```
bskinn, line 53:
```suggestion
        : 'Quansight Labs logo';
```

package-lock.json
Surafel-Getachew: I think the changes happened when I install the dependencies. It's now fixed.

apps/consulting/pages/post/[slug].tsx
```diff
@@ -47,6 +48,11 @@ const Article: FC<TLibraryArticleProps> = ({
         description={post.meta.description}
         variant={DomainVariant.Quansight}
       />
+      <SocialCard
+        title={post.meta.title}
+        description={post.meta.description}
+        variant={DomainVariant.Quansight}
+      />
```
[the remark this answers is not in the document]
Comment (bskinn) [reply apps/consulting/pages/post/[slug].tsx:55]: > I'm guessing we'll have to do more work to make sure that the images are the right dimensions for Twitter and Open Graph

Exactly this, Twitter is 1:1, OpenGraph is 1.91:1. I doubt either exactly matches the ratio for the `featuredImage`; and, for some posts, we will want to be able to specify images of the correct ratio.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+      : fallBackOrigin;
+  const url = `${origin}${asPath}`;
+
+  const twitterCardImg = (): string => {
+    if (twitterCardImgUrl) {
+      return twitterCardImgUrl;
+    } else {
+      return DomainVariant[variant] === 'Quansight'
+        ? 'https://a.storyblok.com/f/147759/1200x1200/86c1d25b6a/quansight-logo-twittersquare.png'
+        : 'https://a.storyblok.com/f/152463/700x700/ea6ce93c47/qslabs-logo-twittersquare.png';
```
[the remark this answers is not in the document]
bskinn: Twitter doesn't use OG, it has its own special protocol.

We looked at this, the Twitter images need to be 1:1, and an OpenGraph viewer isn't going to be helpful.

Apparently the official way to preview social cards on Twitter is ... to draft a tweet containing the link.

And, if you're stuck with a cached version of the social preview image, add a meaningless query to the link, that'll bust the cache.

😖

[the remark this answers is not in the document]
gabalafou: I'm not quite sure what you're saying here, but I would hope that the Vercel engineers are capable of creating an accurate Twitter card preview...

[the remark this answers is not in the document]
gabalafou: I mean, I realize I called it the Vercel open graph inspector, but if you look at the image, Vercel has previews for multiple different platforms, so I think either I'm missing something or maybe you missed something.

apps/consulting/pages/post/[slug].tsx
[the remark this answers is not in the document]
Comment (gabalafou) [reply apps/consulting/pages/post/[slug].tsx:55]: I think both Storyblok and Vercel have image resizing/cropping features... we might be able to use them here... but will take some looking into

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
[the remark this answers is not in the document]
gabalafou: Weird, the Vercel inspector seems to pick up the Twitter-specific set of metadata, including `twitter:card`, but incorrectly renders the preview:

https://vercel.com/quansight/consulting/BnNbkjroiJS87VGP3Z56TF4bZG7o/og

So I guess the Vercel engineers don't know how to create an accurate Twitter card preview, or maybe Twitter has been changing too much recently, I don't know 🤷‍♂️. That's annoying.

I think the preview that Vercel renders actually is a kind of Twitter card, but I think it's the "summary_large_image" type and not the "summary" type that we've specified.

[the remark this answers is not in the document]
bskinn: > I think the preview that Vercel renders actually is a kind of Twitter card, but I think it's the "summary_large_image" type and not the "summary" type that we've specified.

Agreed.

That, potentially, is another knob we could add --- which _type_ of Twitter card to create for a page. I'd recommend that be in a future ticket, though.

Surafel-Getachew:
What I found now is twitter use two types of cards, Summary Card with Large Image and the other we’re using now but in my opinion the card that has large image looks better and gives consistency as the large image summary card looks similar to the one in LinkedIn.
I think what makes the difference is the large one uses 2:1 aspect ration.
Here is the link for more detail
https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary#:~:text=A%20URL%20to,is%20not%20supported

bskinn:
And here's the info page for the large-image variant: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image

Actually, @Surafel-Getachew -- it looks like the only real difference between the two, in terms of metadata, is the `summary` vs `summary_large_image` value for `twitter:card`. The `twitter:creator` tag looks like it's optional.

Let's implement a toggle prop for `SocialCard` to choose between the two types, with the default being the large-image variant. It'll then be up to the content creator to craft the image with the right aspect ratio.

Say: 

```
{/* Large Twitter variant */}
<SocialCard ... />

{/* Small Twitter variant */}
<SocialCard small-twitter ... />
```

Surafel-Getachew:
> summary_large_image

I agree, So we need to update the current twitter default image url to 2:1 aspect ratio.

bskinn:
Twitter appears to be working correctly for both sites--rendering the large-variant cards in both cases.

For LinkedIn, links from Labs are now rendering OK with the LinkedIn Post Inspector, but links from the Consulting preview build are returning a 404.

Given the presence of a Vercel build slug in the canonical URL reported by the Inspector, I'm guessing something is wrong with the new Next.js machinery for injecting the page location:

image

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -18,7 +18,16 @@ export const SocialCard: FC<TSocialCardProps> = ({
 }) => {
   const { asPath } = useRouter();
 
-  const url = `${process.env['NEXT_PUBLIC_VERCEL_URL']}${asPath}`;
```
bskinn: Instead of replacing the fallback machinery entirely, would it work to use `process.env['NEXT_PUBLIC_VERCEL_URL']` instead of `window.location.origin`?

That's more what I was expecting -- pulling Next's URL instead of the window location.

What you had before might have been close to working -- it looks like it was just missing the leading `https://`.

Surafel-Getachew: But I think process.env['NEXT_PUBLIC_VERCEL_URL'] is already contains` https://`

bskinn: Hm. Do you know, @gabalafou?

When we saw [this spurious injection of the domain into the URL](https://github.com/Quansight/Quansight-website/pull/733#issuecomment-1570775167) before, it was because the `https://` scheme was missing from the front of `url`.

bskinn:
@Surafel-Getachew -- while we're working on ironing this last bit out, one thing you can do is to add documentation to the props and internally-defined consts in `SocialCard.tsx`. As you add to and modify the codebase, that's going to be something we'll want you to do consistently, is add/improve documentation.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
Surafel-Getachew: @bskinn seems like adding` https://` works

Surafel-Getachew:
@bskinn Do we have a specific location where we are documenting components.

bskinn:
> bskinn Do we have a specific location where we are documenting components.

No, I was picturing documenting them [JSDoc style](https://jsdoc.app/).

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
gabalafou: Per the docs, [VERCEL_URL does not include https://](https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables), so neither does `NEXT_PUBLIC_VERCEL_URL`. 

Sorry I missed that in my suggestion diff.

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,33 @@
+// og:url -  The canonical URL of your object that will be used as its permanent ID in the graph,
+// example: https://labs.quansight.org/projects
+// example: https://quansight.com/
+// note: should it include query string params?
+// like https://quansight.com/library?page=2&type=blog (?)
```
gabalafou, line 5: I did not intend you to copy these two lines verbatim 😊, sorry for the confusion 

These are questions. I'm not sure what the answer is. But whatever the answer is, it should be documented.

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,33 @@
...
+// 2:1 with minimum dimensions of 300x157 or maximum of 4096x4096 pixels.
+// Images must be less than 5MB in size. JPG, PNG, WEBP and GIF formats are supported.
+// Only the first frame of an animated GIF will be used.
+// SVG is not supported.
+
+// twitter:image for small summary cards
+// aspect ratio of 1:1 with minimum dimensions of 144x144 or maximum of 4096x4096 pixels.
+// The image will be cropped to a square on all platforms.
+// other specification's are similar to large summary cards
+
```
gabalafou, line 22: I think these comment lines should be placed immediately before the type definition and after the import.

Actually, I would break out these comments and match them to each field in the type definition. But that's just my preference.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+    switch (variant) {
+      case DomainVariant.Quansight:
+        return 'Quansight logo';
+      case DomainVariant.Labs:
+        return 'Quansight Labs logo';
+      default:
+        return '';
+    }
+  };
+  console.log('url', url);
```
gabalafou, line 53:
```suggestion
```

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { TSocialCardProps } from './types';
+
+export const SocialCard: FC<TSocialCardProps> = ({
+  title,
+  description,
+  variant,
+  twitterImage,
+  ogImage,
+  alt,
+  summary_large_image = true,
```
gabalafou, line 17:
```suggestion
  summaryLargeImage = true,
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,33 @@
+// og:url -  The canonical URL of your object that will be used as its permanent ID in the graph,
+// example: https://labs.quansight.org/projects
+// example: https://quansight.com/
+// note: should it include query string params?
+// like https://quansight.com/library?page=2&type=blog (?)
+// url?: string;
```
gabalafou, line 6:
```suggestion
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,33 @@
...
+// aspect ratio: The recommended size for an OG Image is 1.91:1. The recommended pixel dimensions of 1200:630 px
+// example: 'https://a.storyblok.com/f/152463/900x472/b289c484e3/qslabs-logo-opengraph-size.png';
+
+// twitter:image for large summary cards
+// 2:1 with minimum dimensions of 300x157 or maximum of 4096x4096 pixels.
+// Images must be less than 5MB in size. JPG, PNG, WEBP and GIF formats are supported.
+// Only the first frame of an animated GIF will be used.
+// SVG is not supported.
+
+// twitter:image for small summary cards
```
gabalafou, line 18: Can it accept relative URLs or should they be absolute?

gabalafou (changes requested):
Thanks for updating! This is coming along! 🚀

libs/shared/ui-components/src/SocialCard/types.ts
bskinn: Agreed, I would also prefer the documentation for each field/prop immediately precede each pro.

libs/shared/ui-components/src/SocialCard/types.ts
bskinn: Thanks for pointing this out, @gabalafou -- I believe it should *not* include query parameters.

The reason I think this is, the main way to bust Twitter's cache for the preview card for a link is to add spurious, new query parameters to that link when composing a test tweet.

If the query parameters mattered to the preview, then this would not be the recommended or observed behavior.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+        content={summary_large_image ? 'summary_large_image' : 'summary'}
+      />
+      <meta name="twitter:site" content="@quansightai" />
+      <meta name="twitter:title" content={title} />
+      <meta name="twitter:description" content={description} />
+      <meta
+        name="twitter:image"
+        content={twitterImage || defaultTwitterImage(variant)}
+      />
+      <meta name="twitter:alt" content={alt || defaultAlt(variant)} />
```
bskinn, line 68: While `twitter:image` can be defined purely depending on whether `twitterImage` is defined or not, I think the logic for `twitter:alt` needs to be more complex.

What happens if someone sets `twitterImage` but doesn't set `alt`? The default alt-text would likely not match the custom image.

I see four cases for how `twitter:alt` should be set:

|              |  `twitterImage` set  |    `twitterImage` unset       |
| :-----:      |       :-------:      |    :-----:                    |
|  `alt` set   |     Use `alt`        |     Use `alt`                 |
|  `alt` unset |    Use empty string  |    Use `defaultAlt(variant)`  |

So, I think something like the following would be better.

```suggestion
      <meta name="twitter:alt" content={alt || (twitterImage ? '' : defaultAlt(variant)}) />
```

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+      {/* open-graph / LinkedIn Specification */}
+      <meta property="og:title" content={title} />
+      <meta property="og:description" content={description} />
+      <meta property="og:url" content={url} />
+      <meta property="og:image" content={ogImage || defaultOgImage(variant)} />
+      <meta property="og:image:alt" content={alt || defaultAlt(variant)} />
```
bskinn, line 75: I think the same `alt` handling here is needed:

```suggestion
      <meta property="og:image:alt" content={alt || (ogImage ? '' : defaultAlt(variant)}) />
```

bskinn (changes requested):

libs/shared/ui-components/src/SocialCard/types.ts
Surafel-Getachew:
```suggestion
export type TSocialCardProps = {
  title: string;
  description: string;
  variant: DomainVariant;
 // twitter:image for small summary cards
 // aspect ratio of 1:1 with minimum dimensions of 144x144 or maximum of 4096x4096 pixels.
 // The image will be cropped to a square on all platforms.
 // other specification's are similar to large summary cards
  twitterImage?: string; 
  // og:url -  The canonical URL of your object that will be used as its permanent ID in the graph,
 // example: https://labs.quansight.org/projects
// example: https://quansight.com/
// note: should it include query string params?
// like https://quansight.com/library?page=2&type=blog (?)
  ogImage?: string;
  alt?: string;
 // twitter:image for large summary cards
// 2:1 with minimum dimensions of 300x157 or maximum of 4096x4096 pixels.
// Images must be less than 5MB in size. JPG, PNG, WEBP and GIF formats are supported.
// Only the first frame of an animated GIF will be used.
// SVG is not supported.
  summaryLargeImage?: boolean;
};

```
My concern is adding the comment just before the types makes it more unreadable. But my suggestion is adding a one line highlighted linked that contains all the info that is add and for information that's not included in the link adding it just before.

Also once we have the documentation updating it to the doc link.

bskinn: > My concern is adding the comment just before the types makes it more unreadable.

Hm -- I use VS Code, and the color formatting that VS Code applies makes it easily readable for me -- easy to tell what's comment and what's code -- and it's easier to associate the comments with the code bits when they're interleaved.

I can see the readability question in the absence of color formatting. But, I think the convention is to put the documentation for each thing immediately above that thing.

@gabalafou, thoughts?

>  But my suggestion is adding a one line highlighted linked that contains all the info that is add and for information that's not included in the link adding it just before.

Can you rephrase this? I'm not sure what you mean. What's "highlighted linked"... just linking to the specification, rather than including all the details in the documentation comment?

bskinn: Also, I think this is the wrong piece of documentation for `ogImage`... doesn't `ogImage` become `og:image`, not `og:url`?

```
// og:url -  The canonical URL of your object that will be used as its permanent ID in the graph,
// example: https://labs.quansight.org/projects
  ogImage?: string;
```

Aren't `og:url` and `twitter:url`

bskinn: Yes, I think it's worth both linking to the original specifications *and* repeating the details in the documentation.

Having the information right there will make it faster for people to work with the component; and, the link will help people if the spec ever changes and the component starts behaving strangely.

Surafel-Getachew: By highlighted link I mean the option you get when you highlight a text and I think it helps because when you open the link it will only highlight the text you want the shared user to focus on

Screenshot 2023-06-01 at 9 18 41 PM

bskinn: Ohh, I see what you mean now.

I would be okay with a URL with a fragment (e.g., `https://example.com/page#go-to-id`) that gets as close to the target content as possible, but I strongly dislike those URLs with highlights. They are *extremely* long, and are especially hard to copy out of GitHub's code views.

Also, I personally find the highlighting more intrusive than helpful. They're also not supported yet by all browsers (e.g., [Firefox](https://connect.mozilla.org/t5/ideas/copy-link-to-highlight/idi-p/16058)).

apps/consulting/pages/post/[slug].tsx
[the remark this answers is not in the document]
Comment (bskinn) [reply apps/consulting/pages/post/[slug].tsx:55]: Following up here; we're implementing so that we have the option for either the 1:1 `summary` Twitter card or the 1.9:1 `summary_large_image` card.

Thus, we'll have to also add metadata to indicate which type of Twitter card is desired.

bskinn:
Ah, yeah - `develop` is the trunk branch, not `main` 🙂

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,71 @@
+import { DomainVariant } from '@quansight/shared/types';
+
+export type TSocialCardProps = {
+  /* 
+  A concise title for the related content.
+  Platform specific behaviors for twitter:
+   iOS, Android: Truncated to two lines in timeline and expanded Tweet
```
bskinn, line 7: Also, please fix the indentation so that each line starts on the same column, within each comment.

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  description: string;
+
+  /**
+   Uses DomainVariant enum values
+   If the optional properties are not added
+   The component uses images and alt based on the variant value
+   */
+  variant: DomainVariant;
+
+  /**
```
bskinn, line 28:
```suggestion
  /**  
  The `summaryLargeImage` prop controls whether the large or small Twitter summary card
  will be used.
  
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,71 @@
...
+    Maximum 420 characters for twitter
+    but for Open Graph there’s no official guidance on this, but 40 characters for mobile and 60 for desktop is roughly the sweet spot.
+    More info: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image
+*/
+  alt?: string;
+
+  /** 
+   The summaryLargeImage accepts a boolean true or false value and makes use of summary_large_image in the name meta tag 
+   eg. <meta name="twitter:card" content="summary_large_image">
+   The component defaults to true value if not specified with value
```
bskinn, line 63:
```suggestion
   The component defaults to true value if a value is not specified here.
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,71 @@
...
+   Remeber to use the twitter:image that fits aspect ration of 2:1 when summaryLargeImage is true
```
bskinn, line 64:
```suggestion
   Remember to use the twitter:image that fits aspect ratio of 2:1 when summaryLargeImage is true,
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,71 @@
...
+   with minimum dimensions of 300x157 or maximum of 4096x4096 pixels.(see twitterImg type above for more detial)
```
bskinn, line 65:
```suggestion
   with minimum dimensions of 300x157 or maximum of 4096x4096 pixels.(see twitterImage type above for more details)
```

bskinn (changes requested):
Making good progress! I think we're getting close.

One functional change still to merge, plus some iteration and fixes on the documentation.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -64,17 +73,20 @@ export const SocialCard: FC<TSocialCardProps> = ({
         name="twitter:image"
         content={twitterImage || defaultTwitterImage(variant)}
       />
+      {/**
```
bskinn, line 76: Ah, sorry, I wasn't clear.

JSDoc, with the `/**`, is only used for documenting objects/props/etc. If, at some point in the future, we set up a full documentation system, the comments starting with `/**` will tell the documentation tool to treat that comment specially, as documentation for the object it precedes.

For other, general code comments, they should have the standard `/*` opening, to indicate they're not meant to appear in the rendered documentation.

```suggestion
      {/*
```

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -64,17 +73,20 @@ export const SocialCard: FC<TSocialCardProps> = ({
...
+       * If `twitterImage` is specified but `alt` is not, we use blank alt-text because
+       * it's better to have no alt-text than incorrect alt-text.
+       */}
       <meta
         name="twitter:alt"
         content={alt || twitterImage ? '' : defaultAlt(variant)}
       />
 
-      {/* open-graph / LinkedIn Specification */}
+      {/** open-graph / LinkedIn Specification */}
```
bskinn, line 85:
```suggestion
      {/* open-graph / LinkedIn Specification */}
```

bskinn (changes requested):
Sorry for not being clearer on this!

bskinn:
@gabalafou, I think this is ready for your re-review.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -18,6 +18,15 @@ export const SocialCard: FC<TSocialCardProps> = ({
...
 
+  /**
+   * url is constructed starting from https:// because 'NEXT_PUBLIC_VERCEL_URL' doesn't provide the protocol(https)
+   * and for og:url requires absolute URL which includes the protocol and domain name
+   * using the prefix NEXT_PUBLIC is used to indicate that the environment variable is public,
+   * meaning that it can be accessed by the browser.
+   * 'NEXT_PUBLIC_VERCEL_URL' is set by Vercel when the application is deployed, and it can be accessed by the application code.
+   * This kind of URL is need because of og:url needs absolute path and allows it to use this as the correct canonical tag
+   * Also twitter doesn't have something similar to og:url like twitter:url.
+   */
```
Surafel-Getachew, line 29: @bskinn What about this comments should I remove the /** or do we need it in the doc. What do you think?

bskinn: I would convert this to `/*`, because it's a comment on an internal variable, not part of the external API of the component. (A user of the component doesn't need to know about how `url` is constructed; this comment is here to help a future developer who is looking into the innards of the component to see how it works.

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Platform specific behaviors for twitter:
+   *  - iOS, Android: Not displayed
+   *  - Web: Truncated to three lines in timeline and expanded Tweet
+   * More info: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image
+   */
+  description: string;
+
+  /**
+   * Uses DomainVariant enum values if the optional properties are not added
+   * SocialCard component uses it to set images and alt based on the variant value.
```
bskinn, line 24:
```suggestion
   * SocialCard component uses it to set default images and alt based on the variant value.
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * twitter:image for large summary cards:
+   * absolute URL must be used rather than relative URL.
+   * 2:1 with minimum dimensions of 300x157 or maximum of 4096x4096 pixels.
+   * Images must be less than 5MB in size. JPG, PNG, WEBP and GIF formats are supported.
+   * Only the first frame of an animated GIF will be used.
+   * SVG is not supported.
+   * more info: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image.
+   * twitter:image for small summary cards:
```
bskinn, line 36:
```suggestion

   * twitter:image for small summary cards:
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * twitter:image for large summary cards:
+   * absolute URL must be used rather than relative URL.
+   * 2:1 with minimum dimensions of 300x157 or maximum of 4096x4096 pixels.
+   * Images must be less than 5MB in size. JPG, PNG, WEBP and GIF formats are supported.
+   * Only the first frame of an animated GIF will be used.
+   * SVG is not supported.
+   * more info: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image.
+
```
bskinn, line 36:
```suggestion
  *
```

bskinn (approved):
This looks good to me.

@gabalafou, we are awaiting your final review and approval for this before merge.

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,71 @@
...
+   * A text description of the image conveying the essential nature of an image to users who are visually impaired.
+   * Maximum 420 characters for twitter.
+   * but for Open Graph there’s no official guidance on this, but 40 characters for mobile and 60 for desktop is roughly the sweet spot.
+   * More info: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image
+   */
+  alt?: string;
+
+  /**
+   * The `summaryLargeImage` prop controls whether the large or small Twitter summary card will be used.
+   * The component defaults to true value if a value is not specified here.
```
bskinn, line 63:
```suggestion
   * The `SocialCard` component *should* be configured with a `true` default value for this prop.
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,71 @@
...
+   * but for Open Graph there’s no official guidance on this, but 40 characters for mobile and 60 for desktop is roughly the sweet spot.
+   * More info: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image
+   */
+  alt?: string;
+
+  /**
+   * The `summaryLargeImage` prop controls whether the large or small Twitter summary card will be used.
+   * The `SocialCard` component *should* be configured with a `true` default value for this prop.
+   * Remember to use the twitter:image that fits aspect ratio of 2:1 when summaryLargeImage is true.
+   * with minimum dimensions of 300x157 or maximum of 4096x4096 pixels.(see twitterImage type above for more details).
```
bskinn, line 65:
```suggestion
   * with minimum dimensions of 300x157 or maximum of 4096x4096 pixels (see twitterImage type above for more details).
```

apps/consulting/pages/post/[slug].tsx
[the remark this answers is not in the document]
Comment (bskinn) [reply apps/consulting/pages/post/[slug].tsx:55]: Created #738

libs/shared/ui-components/src/SocialCard/types.ts
bskinn: Per [here](https://www.everywheremarketer.com/blog/ultimate-guide-to-social-meta-tags-open-graph-and-twitter-cards), if the query parameters can affect which page is shown, you should *include* them.

So, since the query parameters for `/library` on Consulting and `/blog` *do* affect the contents of the page (even if not the social card metadata), it will likely be safer to *include* the query params in `url`.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,96 @@
...
+    }
+  };
+  return (
+    <Head>
+      {/* twitter */}
+      <meta
+        name="twitter:card"
+        content={summaryLargeImage ? 'summary_large_image' : 'summary'}
+      />
+      <meta name="twitter:site" content="@quansightai" />
```
gabalafou, line 69: This doesn't change whether it's Labs or just Quansight? I actually have no idea. Just asking in case there's an oversight here.

gabalafou (changes requested):
Didn't test, just reviewed the code.

I think there's only one thing I caught that needs changing: we should not provide an empty string for the alt text when we simply don't have it. 

All of the my other comments are optional, non-blocking.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
bskinn: No, it should be the same, at least for now. Labs doesn't currently have an independent Twitter presence.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,96 @@
...
+      />
+      {/*
+        If `twitterImage` is specified but `alt` is not, we use blank alt-text because
+        it's better to have no alt-text than incorrect alt-text.
+       */}
+      <meta
+        name="twitter:alt"
+        content={alt || twitterImage ? '' : defaultAlt(variant)}
+      />
+
```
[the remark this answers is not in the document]
bskinn: I think is missing a closing curly brace

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -0,0 +1,96 @@
...
+  summaryLargeImage = true,
+}) => {
+  const { asPath } = useRouter();
+
+  /*
+    url is constructed starting from https:// because 'NEXT_PUBLIC_VERCEL_URL' doesn't provide the protocol(https)
+    and for og:url requires absolute URL which includes the protocol and domain name
+    using the prefix NEXT_PUBLIC is used to indicate that the environment variable is public,
+    meaning that it can be accessed by the browser.
+    'NEXT_PUBLIC_VERCEL_URL' is set by Vercel when the application is deployed, and it can be accessed by the application code.
```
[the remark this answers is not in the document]
Surafel-Getachew: I think that is well written so let's use that.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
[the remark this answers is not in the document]
Surafel-Getachew: @bskinn @gabalafou 

For the suggested code [above](https://github.com/Quansight/Quansight-website/pull/733#discussion_r1214743626)
consider the case when we have an `alt` but no `twitterImage` then what will be shown is a Quansight logo or Labs logo with the wrong `alt`. I believe if we don't have twitterImage we should always show the defaultAlt because we're showing  Quansight logo or Labs logo if we don't have one any way.

to summarize:

| | `twitterImage` | no `twitterImage` |
| -- | ---- | ---------- |
| `alt` | show `alt` | `defaultAlt` |
| no `alt`| don't have `alt` tag | `defaultAlt`|


```suggestion
     {(alt || !twitterImage) && (
        <meta
          name="twitter:alt"
          content={twitterImage ? alt : defaultAlt(variant)}
        />
      )}

```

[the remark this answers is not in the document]
gabalafou: Yeah, that looks right to me.

However, I'm starting to think that the fact that it's taken us this many cycles to nail down this logic suggests that there's something wrong or unintuitive about the API for this React component.

Maybe we got a little bit ahead of ourselves. As far as I can tell, we're not even using these other properties anywhere at the moment. Maybe we should have started with the smallest thing first, just the title, description, and variant props, and added the other props later when we have code that uses them. But since we're already here...

My understanding is that with this component we're basically trying to support two use cases.

1. Social card with generic, default image (Quansight or Quansight Labs logo)
2. Social card with a content-specific image

The first case is easy; it's the one we see in this PR:

```tsx
<SocialCard title="..." description="..." variant={DomainVariant[...]} />
```

The second case is a bit more complicated. But I'm guessing that 90%+ of the time it should look like this:

```tsx
<SocialCard
  title="..."
  description="..."
  variant={...}
  twitterImage="https://..."
  {/* ogImage same as twitterImage, just different size */}
  ogImage="https://..." 
  alt="same alt text for both images because they're both the same, just different sizes"
/>
```

Everything else is an edge case right?

So maybe the simplest implementation for now is to not support any of the edge cases. Which we can accomplish by changing the type:

```ts
type TSocialCardPropsBase = {
  title: string;
  description: string;
  variant: DomainVariant;
};

type TSocialCardPropsDefaultImage = TSocialCardPropsBase;

// If you supply a customized social card image, you must supply the same image
// for both Twitter and Open Graph, and it must have alt text.
type TSocialCardPropsCustomizedImage = TSocialCardPropsBase & {
  twitterImage: string;
  ogImage: string;
  alt: string;
};

export type TSocialCardProps = 
  TSocialCardPropsDefaultImage |
  TSocialCardPropsCustomizedImage;
```

Then you can write the body of the React component code like so:

```tsx
      <meta
        name="twitter:image"
        content={twitterImage || defaultTwitterImage(variant)}
      />
      <meta
        name="twitter:alt"
        content={alt || defaultAlt(variant)}
      />
      <meta property="og:image" content={ogImage || defaultOgImage(variant)} />
      <meta
        property="og:image:alt"
        content={alt || defaultAlt(variant)}
      />
```

You can use simplified logic because the type-checker assures that you're either in the first case using default images and alt text, or you're in the second case where you're using custom images (both Twitter and Open Graph) and alt text.

I don't really know if we need to support the case where somebody provides a Twitter image but not alt text (at Quansight we want to be disciplined about providing alt text). Or the case where somebody provides alt text but no Twitter image—except perhaps in the case where somebody has an image and alt text for Open Graph but not Twitter?  But do we really need to support that? If we want to support the use case where somebody provides a custom image (plus alt text) for Open Graph but not Twitter, or vice versa, then that should probably be reflected in our React component's API. In other words, it should define both `twitterAlt` and `ogImageAlt`.

Thoughts?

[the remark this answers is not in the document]
bskinn: I like your proposed refactoring of the types, @gabalafou. I agree, that either all of these should be present together, or absent together. I originally was thinking about this differently, but I'm aligned with you now.

One thing to add, though: `TSocialCardPropsCustomizedImage` will need an additional e.g. `usingTwitterLargeImage: boolean;` property, to allow the user of the component to indicate which size of twitter card image that `twitterImage` meets.

(The statement `{/* ogImage same as twitterImage, just different size */}` is not correct when using the small-image `summary` Twitter card type.)

---

> we're not even using these other properties anywhere at the moment

No, but the next step I'd planned after this PR merges was to implement use of them elsewhere: to allow custom social card props on blog posts and general site pages.

---

In terms of future code changes:

- Blog posts
  - Change to types for the two sites:
    - Labs: Add a new `socialcard?:` to [`TPostMeta`][labs tpostmeta] in `/apps/labs/types/storyblok/bloks/posts.ts`
    - Consulting: Add a new `socialcard?:` to [`TPostMeta`][cons tpostmeta] in `/apps/consulting/types/storyblok/bloks/blogPost.ts`
    - More inconsistency between the two sites, in this filename 🫤
  - Because of these types that we can enforce, a failure to provide all four of the needed metadata fields in a post header should lead to a broken build.
- Regular (Storyblok) site pages
  - The SB schema for `Page` will need to be updated with these four fields.
  - I _think_ any validation to require that all four fields be present or absent together will have to happen on the Storyblok side.
    - The page metadata is [pulled from the page's `data.content`][data.content], which is pulled from the Apollo query.
    - Given that, e.g., `data.content.description` is optional in Storyblok and there's no indication I can see of its being optional in [the Apollo query], I don't think we can enforce this on the codebase side.
    - So... one option might be to implement a `SocialCard` component in Storyblok? Where its presence on a `Page` is optional, but all four fields within it are required?


[cons tpostmeta]: https://github.com/Quansight/Quansight-website/blob/674832b40fac9f17f54daa566642ab3c238e7ad6/apps/consulting/types/storyblok/bloks/blogPost.ts#L19-L29
[labs tpostmeta]: https://github.com/Quansight/Quansight-website/blob/05a6d3c818017eca3f2186233275ce27e98580d9/apps/labs/types/storyblok/bloks/posts.ts#L28-L38

[data.content]: https://github.com/Quansight/Quansight-website/blob/674832b40fac9f17f54daa566642ab3c238e7ad6/apps/consulting/pages/index.tsx#L33-L37

[the apollo query]: https://github.com/Quansight/Quansight-website/blob/674832b40fac9f17f54daa566642ab3c238e7ad6/libs/shared/storyblok-sdk/src/api/queries/common/page.graphql#L12-L20

bskinn (commented):
All makes sense.

@Surafel-Getachew, I'd say if you agree with any of my recommended changes, please go ahead and make them.

Any you're unsure of or disagree with, we can leave for @gabalafou to evaluate.

Thanks for this!

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -68,4 +74,8 @@ export interface TSocialCardProps {
    * More info: `https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image`
    */
   summaryLargeImage?: boolean;
```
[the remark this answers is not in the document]
Surafel-Getachew: let's wait @gabalafou. I'm unsure about this.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -5,25 +5,34 @@ import { useRouter } from 'next/router';
...
+
+  const { title, description, variant } = props;
+
   /*
     url is constructed starting from https:// because 'NEXT_PUBLIC_VERCEL_URL' doesn't provide the protocol(https)
     and for og:url requires absolute URL which includes the protocol and domain name
     using the prefix NEXT_PUBLIC is used to indicate that the environment variable is public,
     meaning that it can be accessed by the browser.
-    'NEXT_PUBLIC_VERCEL_URL' is set by Vercel when the application is deployed, and it can be accessed by the application code.
+    'NEXT_PUBLIC_VERCEL_URL' is set when the site is statically built, the build step inline the current value for the environment variable into the built JS file.
```
gabalafou, line 35:
```suggestion
    'NEXT_PUBLIC_VERCEL_URL' is set when the site is statically built, the build step inlines the current value for the environment variable into the built JS file.
```

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -40,32 +46,36 @@ export interface TSocialCardProps {
...
-   * Remember to use the twitter:image that fits aspect ratio of 2:1 when summaryLargeImage is true.
+   * Remember to use the twitter:image that fits aspect ratio of 2:1 when twitterLargeImage is true.
    * with minimum dimensions of 300x157 or maximum of 4096x4096 pixels (see twitterImage type above for more details).
    * full-width prominent image alongside a tweet. It is designed to give the reader a rich photo experience,
    * and clicking on the image brings the user to your website.
    * More info: `https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image`
    */
-  summaryLargeImage?: boolean;
-}
+  twitterLargeImage?: boolean;
```
gabalafou, line 76: I think this should be an optional field of the base type.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
```diff
@@ -59,37 +68,41 @@ export const SocialCard: FC<TSocialCardProps> = ({
...
   return (
     <Head>
       {/* twitter */}
       <meta
         name="twitter:card"
-        content={summaryLargeImage ? 'summary_large_image' : 'summary'}
+        content={
+          isCustom
+            ? getTwitterCardTypeName(props.twitterLargeImage)
+            : getTwitterCardTypeName()
```
gabalafou, line 80: Could this be rewritten as

```suggestion
          twitterLargeImage 
            ? TWITTER_LARGE_CARD_NAME 
            : TWITTER_SMALL_CARD_NAME;
```

gabalafou (approved):
I have a few suggestions, but not really blocking

libs/shared/ui-components/src/SocialCard/types.ts
bskinn: Ah, interesting -- and then we would need to have default images for both the large and small card formats, switchable by this field.  Agreed.

libs/shared/ui-components/src/SocialCard/SocialCard.tsx
bskinn: If we pull `twitterLargeImage` over to the base type as you suggest, I would think so.

libs/shared/ui-components/src/SocialCard/types.ts
Surafel-Getachew: Making it a base type means someone got to decided to use large or small twitter card even if they don't provide twitter image. I think if what is showing in the card is Quansight or labs logo we should use the large one as it's more expressive.

WDYT?

bskinn: Hm -- if `twitterLargeImage` is optional on the base type, though, that means that a blog author can omit it from the metadata -- in which case the selection would fallback to the default, and I'm fine with that default being the large size.

bskinn:
Once we've finalized the changes here, we will need to update the instructions/documentation in the README to match.

This will also influence the language we can use there. In particular, the suggestion of "All metadata fields are required for all blog posts." in https://github.com/Quansight/Quansight-website/pull/739#issuecomment-1591333038 will not work.

libs/shared/ui-components/src/SocialCard/types.ts
Surafel-Getachew: but that will give the blog author a permission to use small image of Quansight or labs if they set's it `twitterLargeImage` to `false`. if you think that 's ok we need a small image for twitter.

gabalafou: I say let's leave the code as is and merge it in. And if we have a real need to add support for small Twitter images for the default/fallback Quansight and Quansight Labs cards, let's add it later. 

I would prefer if we do not spend any more time shaping this component for use cases that are not actually used (yet). 

More details, if you care to read on ...

I somewhat carelessly made the suggestion to move the `twitterLargeImage` field because the first time I read 

```ts
isCustom
  ? getTwitterCardTypeName(props.twitterLargeImage)
  : getTwitterCardTypeName()
```

I thought, why not just write:

```ts
getTwitterCardTypeName(props.twitterLargeImage)
```

and then I was like, oh right, because TypeScript will complain. But I didn't think through the implication that Surafel pointed out.

For what it's worth, you could move the field to the base type and then write:

```ts
export type TSocialCardPropsDefaultImage = TSocialCardPropsBase & {
  // Our default Twitter cards for Quansight and Quansight Labs are large
  twitterLargeImage?: true;
};
```

Then you could write code in the SocialCards component that looks a bit more natural and less like a TypeScript workaround.

gabalafou (approved):

bskinn:
Me:

> Once we've finalized the changes here, we will need to update the instructions/documentation in the README to match.

Now that I think about this again, I don't think we need to make any README changes for this PR. We only will need to add to the README instructions once we implement the ability to customize the social cards using blog post metadata, and that will be in a future PR. (@Surafel-Getachew, please correct me if I'm wrong here.)

If this is correct, then the PR is nearly ready for merge.

@Surafel-Getachew, it looks like you need to update the `Surafel-Getachew:develop` branch of your fork to match `Quansight-website:develop` of the main repo here, and then either merge `Surafel-Getachew:develop` into `Surafel-Getachew:social-cards` or rebase `Surafel-Getachew:social-cards` onto the tip of `Surafel-Getachew:develop`.

Surafel-Getachew:
> Me:
> 
> > Once we've finalized the changes here, we will need to update the instructions/documentation in the README to match.
> 
> Now that I think about this again, I don't think we need to make any README changes for this PR. We only will need to add to the README instructions once we implement the ability to customize the social cards using blog post metadata, and that will be in a future PR. (@Surafel-Getachew, please correct me if I'm wrong here.)
> 
> If this is correct, then the PR is nearly ready for merge.
> 
> @Surafel-Getachew, it looks like you need to update the `Surafel-Getachew:develop` branch of your fork to match `Quansight-website:develop` of the main repo here, and then either merge `Surafel-Getachew:develop` into `Surafel-Getachew:social-cards` or rebase `Surafel-Getachew:social-cards` onto the tip of `Surafel-Getachew:develop`.

That's right @bskinn as of now the blog author can't customize the social cards using blog post metadata

I think we can merge now.

libs/shared/ui-components/src/SocialCard/types.ts
```diff
@@ -0,0 +1,82 @@
...
+   * The `twitterLargeImage` prop controls whether the large or small Twitter summary card will be used.
+   * The `SocialCard` component *should* be configured with a `true` default value for this prop.
+   * Remember to use the twitter:image that fits aspect ratio of 2:1 when twitterLargeImage is true.
+   * with minimum dimensions of 300x157 or maximum of 4096x4096 pixels (see twitterImage type above for more details).
+   * full-width prominent image alongside a tweet. It is designed to give the reader a rich photo experience,
+   * and clicking on the image brings the user to your website.
+   * More info: `https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image`
+   * Our default Twitter cards for Quansight and Quansight Labs are large
+   */
+  twitterLargeImage?: true;
```
bskinn, line 38: @Surafel-Getachew -- With this moved to the base class, we'll need those 1:1 logo images for the small Twitter images, and to add selection code for them into `SocialCard.tsx:defaultTwitterImage`, right?

bskinn: Or, we could move `twitterLargeImage` back into `TSocialCardPropsCustomizedImage` -- I think this might actually have been what @gabalafou meant to do.

Surafel-Getachew: Ah, @bskinn you are right WDYT should we revert back?

bskinn: Yeah, I would revert that back. Then, I think we're ready to merge.

bskinn:
Social cards for one blog post and one regular page from each of the sites are rendering ok on both Twitter and the LinkedIn post inspector.

In the interest of time, I'm not going to test every possible page from each site. If there are pages with glitches in their social card implementation, they will be discovered in time.

Proceeding with merge.

Issue #403, opened 2022-08-02
[BUG] - Labs and Consulting blog RSS feed generation is broken, and disabled


bskinn:
### What site is this for?

Quansight Labs
Quansight Consulting

### Expected behavior

RSS feed generation for the blog posts for both sites should occur on every build.

### Actual behavior

Using the Labs site as an example:

The RSS feed is not generated, as the function call to generate it is [commented out](https://github.com/Quansight/Quansight-website/blob/25b970931afdf4d1498ad376d91e355fe4e7e0a3/apps/labs/pages/blog/index.tsx#L211).

Upon testing, uncommenting the generation function [causes the Labs site build to break](https://vercel.com/quansight/quansight-labs-live/71aooYpEezRdfL2ppTVcBkk1V66p).

This needs to be fixed. ~We will have to decide whether we also want an RSS feed generated for the LLC blog, as part of #396.~ Yes, we desire this for Consulting blog also.

bskinn:
Added the LLC label since this will apply to LLC blog after #396 is complete.

bskinn:
#739 is addressing the broken RSS feed for Labs.

Further discussion on the approach to expanding implementation of the RSS feed for Consulting blog, prior to actually starting work, should take place here.

bskinn:
I've created #741 to track any discussion on RSS for the Consulting blog, and am closing this one after the merge of #739.

kalvdans:
Sorry for posting a question here, but what is the URL to the rss feed of the Labs blog? There's no autodiscover tag on https://labs.quansight.org/blog .

bskinn:
No problem at all, @kalvdans! That autodiscover tag sounds like something we should add, as well as a direct RSS link in the footer, probably.

The feed is at https://labs.quansight.org/rss.xml.

Thanks for your interest!

PR #402, opened 2022-08-02
Remove 'hello world' items from rss.xml

bskinn:
They keep reappearing on https://planetpython.org/index.html and
are a nuisance to visitors of the site.

Closes #346

Issue #346, opened 2022-07-13
[BUG] - RSS feed is broken, "hello world" posts showing up on Planet Python/SciPy


rgommers:
### What site is this for?

Quansight Labs

### Expected behavior

The RSS feed at https://labs.quansight.org/rss.xml should contain valid content for new blog posts.

### Actual behavior

The current RSS feed is broken, showing dummy posts with today's or yesterday's date. Those are showing up on https://www.planetpython.org/ and https://planet.scipy.org/:

<img width="537" alt="image" src="https://user-images.githubusercontent.com/98330/178655064-ac575100-140f-4d25-9081-5f484a469f35.png">

<img width="537" alt="image" src="https://user-images.githubusercontent.com/98330/178655104-9c033dea-de50-47e2-b711-160f4cc816ed.png">

I was made aware of this through https://github.com/python/planet/issues/517. We should fix this asap, or remove the feeds from the Planet aggregators.

### How to Reproduce the problem?

Looking at https://labs.quansight.org/rss.xml is enough. After the fix, please check with an RSS feed reader that the content is correct.

bskinn:
This should now be fixed, @rgommers. Sorry it took us so long! I'd looked a while ago at the blog posts directory and seen that the 'hello world' posts were gone, and figured it was all sorted. Did not realize that the RSS generation was broken and had been disabled.

PR #739, opened 2023-06-02
[LAUNCH CLEANUP] Restore Labs RSS feed

Surafel-Getachew:
### Background

The code to generate RSS for the Labs site was commented out because, at various points in the past, it was causing (or was thought to be causing) the build to fail.

So we investigated why the build was failing and, as per the [build log error message](https://vercel.com/quansight/quansight-labs-live/71aooYpEezRdfL2ppTVcBkk1V66p) observed at the time (Aug 2022), the `featuredImage` metadata field was missing from one of the in-draft posts at the time.

The specific error message, `Cannot read properties of undefined (reading 'src')`, occurred because the RSS generation code was trying to access `featuredImage.src`, when `featuredImage` was `undefined`.

See https://github.com/Quansight/Quansight-website/pull/739#issuecomment-1585152218 below for more details on the timeline of the thought-to-be-broken status of the RSS generation code.

### Why make this change?

After re-enabling the RSS generation, the Labs build succeeded and generated a well formed `rss.xml`. Also:

- We iterated over all the Labs posts and confirmed that all posts defined `featuredImage` and its child fields.
- We also verified the prior error was due to a missing `featuredImage` property by removing `featuredImage` from a blog post and checking the build. We got the same [build failure](https://vercel.com/quansight/labs/D6VmXWRN7Wq4Esdzrra82knR9wZy) as was observed before.

### How does it relate to issue #403?

#403 has two elements:

- Fix RSS generation for Labs site
- Implement RSS generation for Consulting site

This PR "fixes" the **Labs** RSS feed generation by observing that it likely was not broken in the first place and re-enabling that code pathway.

This PR **does not address** implementation of an RSS feed for the Consulting blog.

### Documentation improvements

One of the main findings from this is that posts have required metadata fields that are needed for the build to not fail. Meta properties for the posts are covered in the [README](https://github.com/Quansight/Quansight-website#structure-of-the-blog-post) -- this PR adds language indicating which fields are required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Here is some basic info to help orient you to this repo.
   - `./libs` holds code shared by both websites.
 - The websites' **content** lives in [Storyblok](https://app.storyblok.com)
   (requires login).
-  - But Labs and Consulting **blog posts** live under `./apps/labs/posts` and
+  - But Labs and Consulting **blog posts** live under `./apps/labs/posts` and 
     `./apps/consulting/posts`, respectively.
 - The websites are hosted and deployed via
   [Vercel](https://vercel.com/quansight) (requires login).
@@ -354,12 +354,12 @@ Throughout the develop-deploy process, there are several ways to view the
 website. The following table summarizes the important ways in which those views
 differ from each other.
 
-| Name                        | How to access          | GitHub branch  | Vercel env | Storyblok API key | Next.js preview? | Storyblok version param | Display top overlay? | Top overlay color | Button to enter/exit preview? |
-| --------------------------- | ---------------------- | -------------- | ---------- | ----------------- | ---------------- | ----------------------- | -------------------- | ----------------- | ----------------------------- |
-| Production                  | .com/.org URL          | `main`         | production | public            | off              | `published`             | No                   | n/a               | n/a                           |
-| Storyblok (yellow overlay)  | via Storyblok UI       | `develop`      | preview    | preview           | on               | `draft`                 | Yes                  | yellow            | No                            |
-| Vercel URL (gray overlay)   | via link to Vercel URL | any non-`main` | preview    | preview           | off              | `published`             | Yes                  | gray              | Yes                           |
-| Vercel URL (yellow overlay) | via enter-preview      | any non-`main` | preview    | preview           | on               | `draft`                 | Yes                  | yellow            | Yes                           |
+| Name                       | How to access          | GitHub branch  | Vercel env | Storyblok API key | Next.js preview? | Storyblok version param | Display top overlay? | Top overlay color | Button to enter/exit preview? |
+| -------------------------- | ---------------------- | -------------- | ---------- | ----------------- | ---------------- | ----------------------- | ------------------- | ---------------- | ----------------------------- |
+| Production                 | .com/.org URL          | `main`         | production | public            | off              | `published`             | No                  | n/a              | n/a                           |
+| Storyblok (yellow overlay)  | via Storyblok UI       | `develop`      | preview    | preview           | on               | `draft`                 | Yes                 | yellow           | No                            |
+| Vercel URL (gray overlay)   | via link to Vercel URL | any non-`main` | preview    | preview           | off              | `published`             | Yes                 | gray             | Yes                           |
+| Vercel URL (yellow overlay) | via enter-preview      | any non-`main` | preview    | preview           | on               | `draft`                 | Yes                 | yellow           | Yes                           |
 
 Let's take the row labeled "Vercel URL (gray overlay)." This view is accessed by
 clicking on a Vercel SHA-style URL, which looks like
--- a/README.md
+++ b/README.md
@@ -547,9 +547,9 @@ signs. The meta section contains post-related information like:
   categories should be the same as in the previously mentioned
   [`categories.json`](./apps/labs/posts/categories.json) file.
   **Important note:** categories are case-sensitive.
-- `featuredImage (required)` - Object with properties: `src` and `alt`. The `src (required)` property
+- `featuredImage (required)` - Object with two required properties: `src` and `alt`. The `src` property
   is a path to the featured image which is displayed on the posts list on the`/blog`
-  page. The `alt (required)` property is alternative text for the image. The image should
+  page. The `alt` property is alternative text for the image. The image should
   be added to the `apps/labs/public/posts/<post-name>` directory, for example,
   `apps/labs/public/posts/hello-world-post`. There is no need to provide a full image path,
   so the path name should start with `/posts/`.
--- a/README.md
+++ b/README.md
@@ -553,7 +553,7 @@ signs. The meta section contains post-related information like:
   be added to the `apps/labs/public/posts/<post-name>` directory, for example,
   `apps/labs/public/posts/hello-world-post`. There is no need to provide a full image path,
   so the path name should start with `/posts/`.
-- `hero` - the object for the Hero section of the post. This can have two different structures:
+- `hero` (required) - the object for the Hero section of the post. This can have two different structures:
   - The first structure is an object with `imageSrc` and `imageAlt`. The `imageSrc` property is a path to
     the hero image, which is displayed on the blog post page between the nav bar and the
     blog heading title. The `imageAlt` property is alternative text for the image.
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Here is some basic info to help orient you to this repo.
   - `./libs` holds code shared by both websites.
 - The websites' **content** lives in [Storyblok](https://app.storyblok.com)
   (requires login).
-  - But Labs and Consulting **blog posts** live under `./apps/labs/posts` and 
+  - But Labs and Consulting **blog posts** live under `./apps/labs/posts` and
     `./apps/consulting/posts`, respectively.
 - The websites are hosted and deployed via
   [Vercel](https://vercel.com/quansight) (requires login).
@@ -354,12 +354,12 @@ Throughout the develop-deploy process, there are several ways to view the
 website. The following table summarizes the important ways in which those views
 differ from each other.
 
-| Name                       | How to access          | GitHub branch  | Vercel env | Storyblok API key | Next.js preview? | Storyblok version param | Display top overlay? | Top overlay color | Button to enter/exit preview? |
-| -------------------------- | ---------------------- | -------------- | ---------- | ----------------- | ---------------- | ----------------------- | ------------------- | ---------------- | ----------------------------- |
-| Production                 | .com/.org URL          | `main`         | production | public            | off              | `published`             | No                  | n/a              | n/a                           |
-| Storyblok (yellow overlay)  | via Storyblok UI       | `develop`      | preview    | preview           | on               | `draft`                 | Yes                 | yellow           | No                            |
-| Vercel URL (gray overlay)   | via link to Vercel URL | any non-`main` | preview    | preview           | off              | `published`             | Yes                 | gray             | Yes                           |
-| Vercel URL (yellow overlay) | via enter-preview      | any non-`main` | preview    | preview           | on               | `draft`                 | Yes                 | yellow           | Yes                           |
+| Name                        | How to access          | GitHub branch  | Vercel env | Storyblok API key | Next.js preview? | Storyblok version param | Display top overlay? | Top overlay color | Button to enter/exit preview? |
+| --------------------------- | ---------------------- | -------------- | ---------- | ----------------- | ---------------- | ----------------------- | -------------------- | ----------------- | ----------------------------- |
+| Production                  | .com/.org URL          | `main`         | production | public            | off              | `published`             | No                   | n/a               | n/a                           |
+| Storyblok (yellow overlay)  | via Storyblok UI       | `develop`      | preview    | preview           | on               | `draft`                 | Yes                  | yellow            | No                            |
+| Vercel URL (gray overlay)   | via link to Vercel URL | any non-`main` | preview    | preview           | off              | `published`             | Yes                  | gray              | Yes                           |
+| Vercel URL (yellow overlay) | via enter-preview      | any non-`main` | preview    | preview           | on               | `draft`                 | Yes                  | yellow            | Yes                           |
 
 Let's take the row labeled "Vercel URL (gray overlay)." This view is accessed by
 clicking on a Vercel SHA-style URL, which looks like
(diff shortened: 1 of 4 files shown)


Surafel-Getachew:
Screenshot 2023-06-03 at 12 29 59 AM

I think the issue happened before may be because we have a post with no `src` property in `post.meta.featuredImage.src` since the log error also says TypeError

I've checked in the preview deployment and its generating correct `rss.xml` https://labs-a5kipk1gu-quansight.vercel.app/rss.xml 

But I'd suggest verifying the issue by removing some of the post attributes like the `src` and checking if the issue persist or if we can find full log of the build that caused the error taking a look at that will also give us hint where exactly the issue happened.

bskinn:
Sounds like you think [this](https://github.com/Quansight/Quansight-website/blob/2146e0475c09984bd29801be1bea7ed63a283636/apps/labs/services/posts/generateRSS.ts#L37) is where the problem is, @Surafel-Getachew -- and I agree, that seems highly likely.

It seems like we have two main options at the outset:

1. Hunt through the posts to try to find one without `featuredImage.src` in the metadata.
2. Modify the RSS generation code to fall back to a default image if either of `featuredImage` or `feauredImage.src` is not present.

(2) would be fast, but ultimately a bandaid. Thus, I'd be inclined to go ahead and do (1) -- the Labs posts live [here](https://github.com/Quansight/Quansight-website/tree/develop/apps/labs/posts) & the metadata is defined in the Markdown files via the [YAML syntax variant](https://stackoverflow.com/a/44222826/4376000).

However, `featuredImage` is used to provide the image for the gallery view on https://labs.quansight.org/blog, and I suspect we would have noticed if any of the posts were missing its gallery image.

So, my guess is that (1) will actually leave us in search of a deeper problem to debug.

----

One thing to keep in mind as we're working on this is that, in addition to fixing the RSS feed for Labs, we're also going to want to implement it for Consulting. So, even though `generateRSS.ts` right now lives in the `/apps/labs` portion of the directory tree, I think ultimately it would make sense to refactor it into the common code of `/libs/shared`.

Surafel-Getachew:
@bskinn just to confirm are we using [this](https://github.com/Quansight/Quansight-website/tree/develop/apps/labs/posts) posts in prod also, because the build is failing in prod

bskinn:
> bskinn just to confirm are we using [this](https://github.com/Quansight/Quansight-website/tree/develop/apps/labs/posts) posts in prod also

Yes, all of those posts are building to the live site.

> because the build is failing in prod

I'm not sure what you mean by this -- the build *is* succeeding on the live site because the RSS generation is disabled (commented out) on `develop`.

Huh. And, actually, the most recent preview build for this branch [completed just fine](https://labs-a5kipk1gu-quansight.vercel.app/).

Is the build failure non-deterministic?

Surafel-Getachew:
@bskinn what I meant to say is issue #403 is because it's failing in prod at least that's what the Vercel build is showing 

Screenshot 2023-06-09 at 9 33 09 PM

and I checked all the posts in here [here](https://github.com/Quansight/Quansight-website/tree/develop/apps/labs/posts) and all posts contain `featuredImg` and `featuredImg.src` and I am wondering if we're getting the posts from a different place

bskinn:
Y'know, I think I completely misread your initial comment (https://github.com/Quansight/Quansight-website/pull/739#issuecomment-1574351437).

> I've checked in the preview deployment and its generating correct rss.xml

You're right, it *is* building ok on the preview.

> it's failing in prod 

Where are you looking for this/what are you calling "prod"? The most recent builds on both [`main`] and [`develop`] both were successful (but RSS generation is disabled for both branches, so they're irrelevant to this issue). If you're clicking the "[causes the Labs site build to break](https://vercel.com/quansight/quansight-labs-live/71aooYpEezRdfL2ppTVcBkk1V66p)." link in #403, that goes to a *very old* build log, not to prod.

----

In any event -- no, there should be no other source for the blog posts to be drawn from. If it's working on this branch, it should work if merged to `develop`. 

I agree -- testing the hypothesis by removing a `featuredImage.src` metadata attribute and seeing if the build breaks in the same way as before is a good idea.

If indeed the problem was only a missing `featuredImage.src` attribute on a post, which has since been fixed, then that's great! The next step will be to evaluate the best way to enable RSS feed generation on *both* the Consulting and Labs sites.

I believe the big question is whether it would be a good idea to refactor the RSS code to a common location in `/libs/shared`, or to implement it separately for Consulting vs Labs. This'll depend on whether there's enough commonality between the two sites.

This discussion of expanding the implementation to Consulting site should take place in the #403 discussion thread, though, not here.

[`main`]: https://vercel.com/quansight/labs/DyT6neLqWA1c4eLEsfB6M4RvVJnh
[`develop`]: https://vercel.com/quansight/labs/7Co482tsqqE9VMezaTYkBoP55qJm

bskinn:
Odd that it still built.

Looking back at the original failed deployment in #403, I used the 'Source' tab in the submenu on the Vercel page to go back to the commit in GitHub that the build was from.

Turns out there was a [post](https://raw.githubusercontent.com/Quansight/Quansight-website/e0c76bbd518be6816909334a940dba1f3b93babb/apps/labs/posts/alt-text-scipy-2022.md) that had `featuredImage` *completely commented out* in its metadata.

So, it might be that it'll build ok if `featuredImage.src` is missing, but breaks if `featuredImage` is missing. That makes sense, because then JS is trying to access a `src` attribute on an `undefined`.

Surafel-Getachew:
@bskinn. that's great. I think that was the problem.

I think the next step todo is:

- [ ] verify the build fails by removing/commenting the `featuredImg` 
- [ ] add the `featuredImg` that was removed for verifying build
- [ ] remove the comment from `generateRSS`
- [ ] document the finding and how it's verified.

WDYT?

bskinn:
I agree, @Surafel-Getachew -- please proceed!

In terms of documentation and verification, the discussion in this PR might suffice for both documentation and verification. @gabalafou, thoughts?

To avoid breakage occurring in the future, we could consider strengthening the language in the [blog post section of the README](https://github.com/Quansight/Quansight-website#structure-of-the-blog-post); indicating that `featuredImage` and its children are _required_ for all posts. Documenting that Again, @gabalafou, thoughts?

I will also document here some of the research I did just now, looking into the history of the codebase around this.

Surafel-Getachew:
https://vercel.com/quansight/labs/D6VmXWRN7Wq4Esdzrra82knR9wZy the build failed when `featuredImage` is removed.

bskinn:
Yep, looks like the same build error to me.

bskinn:
### Tracing the history of interactions with the `generateRSS()` call in `/apps/labs/pages/blog/index.tsx`

* The current state in `develop` a/o 9 Jun 2023 ([58baa57]) has `generateRSS()` commented out.
* The `generateRSS()` call was initially commented out by marcin-antczak-naturaily on 29 Jun 2022 in [c8f7a50] as part of an effort to fix a build problem during the run-up to the site launch.
  * It's unclear at this time if the RSS feed was actually part of the problem or not.
  * Regardless, at this time a stub `rss.xml` file had been created in the repository from a prior call to `generateRSS()` and was carried forward unmodified.
* On 13 Jul 2022, @rgommers reported #346.
  * The stub `rss.xml` file contained two post-dated 'hello world' post items, which were being repeatedly picked up by Planet Python's aggregator.
  * #402 fixed the overt problem by removing these post-dated feed items.
  * In the course of work on `#402`, since `#346` was related to RSS, a test re-enable of `generateRSS()` was attempted (see [e977b8e]), at which time the build broke again.
    * As it was not a critical feature for the site launch, `generateRSS()` was then re-disabled without further investigation and the problem was documented in #403.
* In the course of the above investigation into the RSS build machinery (Jun 2023), @Surafel-Getachew [discovered][success build comment] that re-enabling `generateRSS()` led to a successful build.
  * Subsequent investigation revealed that the build breakage during the work on `#402` almost certainly occurred because, at the time, the post at `/blog/alt-text-scipy-2022` [had its `featuredImage` metadata commented out][no featureimage].
    * With no `featuredImage` attribute present, the attempt to access `featuredImage.src` became an attribute access on `undefined`, which triggered the build error.


### Upshot

The Labs RSS feed capability likely has actually been in good working order for quite some time, and may never have been broken. It seems appropriate to re-enable the feature and call it fixed for Labs.

As noted above, we will want this on Consulting site also, if we return to the Jamstack architecture. It is **low** priority to implement RSS for the Consulting blog at this time, since we do not know if we will return to Jamstack from Wordpress.

[58baa57]: https://github.com/Quansight/Quansight-website/blob/58baa570cb766cf33d65cba71519f4c458dbd677/apps/labs/pages/blog/index.tsx#L217
[c8f7a50]: https://github.com/Quansight/Quansight-website/commit/c8f7a5064661c664f6c264873ce350eb40b4d81b#diff-a4bb3a84b6c71b9870c280550638d24d27d722712f2f6788e22600d7d037cbf8R209
[e977b8e]: https://github.com/Quansight/Quansight-website/pull/402/commits/e977b8eeb30a8607cf0c0be56715287570e3f4ca#diff-a4bb3a84b6c71b9870c280550638d24d27d722712f2f6788e22600d7d037cbf8R209
[success build comment]: https://github.com/Quansight/Quansight-website/pull/739#issuecomment-1574351437
[no featureimage]: https://github.com/Quansight/Quansight-website/blob/e977b8eeb30a8607cf0c0be56715287570e3f4ca/apps/labs/posts/alt-text-scipy-2022.md?plain=1#L7-L9

bskinn:
@Surafel-Getachew Please revert the linter changes to `alt-text-scipy-2022.md` for this PR.

Since the fix doesn't involve that post, better to not have it touched by the PR's diff.

bskinn:
> There's a lot to digest in the conversation on this PR. To be honest, I'm struggling to digest all of it.
> 
> Before merging this PR, its title and description should both be fixed. The title should reflect the actual change made in the code—so something like "Restore Labs RSS feed". The description should explain why you are making this change now, why it's safe now to do so. How does it relate to issue #403? The description should also summarize the key findings from the long conversation on this PR that are relevant to this particular code change.

It may suffice to port https://github.com/Quansight/Quansight-website/pull/739#issuecomment-1585152218 up to the description, except re-ordered to put the "Upshot" section before the bulleted list of details.

> 
> That's why I am marking this PR as "request changes." Once the title and description are fixed, I will be in a better position to approve it.
> 
> As for the questions raised in earlier discussion, yes I think it would be good to update the README to reflect `featuredImage` being required, not optional (preferably with an explanation for why it's required).

It looks to me like _all_ of the [metadata fields](https://github.com/Quansight/Quansight-website/blob/develop/README.md#structure-of-the-blog-post) are required. It might suffice to add a bolded "**All metadata fields are required for all blog posts.**" sentence to the introductory paragraph of that README section?

> That said, I think it's even more important that the build step should fail loudly if `featuredImage` is missing (but required) from a blog post. Fortunately, unless I've misread the discussion on this PR, it seems to me that that is already the case (yay!).

Yes, the build will fail during RSS generation if `featuredImage` is missing.

> 
> Ideally, the build should fail in a way that's clear and not cryptic, but that may be outside the scope of this PR. For example, if I'm trying to add a blog post to Labs and I haven't added a featured image, I should see that the build fails, and it would be nice if the build log indicates that the reason it failed is because of the missing featured image, so that when the blog author pings me or somebody else with access to the logs, it will be easy to diagnose the root cause of the failure and relay that info back to the blog author.

I agree, and I see this as one example of a broader problem with this codebase overall: as best I can tell, there's essentially no error handling at all. I don't think I've seen a single build error that describes well what broke; every time it's been a detective investigation to try to infer where the real problem is.

It seems likely not worth it to try to comb the codebase and pre-emptively add better errors throughout. But, I would be fully in support of a standing policy that we consider adding better error handling and log messages in response to any even mildly cryptic build error.

Given that we are examining these errors in log output, and not working with them programmatically, it would probably suffice to throw undifferentiated `Error`s with well crafted messages, as opposed to standing up a custom error inheritance hierarchy?

> 
> To reiterate, the most important thing is: if I submit a blog post that would break the RSS feed (because it's missing a featured image), then the build should fail on my PR. If that's NOT the case, I don't want to merge this PR until we add guardrails. I think the guardrails are already there, I'm just asking twice to make sure.

Yes, the build will fail in that case, because the RSS feed generation itself will error. But, per above, having an improved error message would be even more valuable.

README.md
```diff
@@ -534,22 +534,22 @@ The `content` section is the body of the post added in Markdown format.
...
+- `author (required)` - Unique slug of the author (from Storyblok) usually looks like `jon-doe`.
   Based on this property blog post page will display proper info about the author
   (and their avatar). The author must be present in Storyblok in order for the
   post to build without error.
 - `category` - Array of categories for example `[Machine Learning]`. All
   categories should be the same as in the previously mentioned
   [`categories.json`](./apps/labs/posts/categories.json) file.
   **Important note:** categories are case-sensitive.
-- `featuredImage` - Object with properties: `src` and `alt`. The `src` property
+- `featuredImage (required)` - Object with properties: `src` and `alt`. The `src (required)` property
```
bskinn, line 550: I think it's cleaner to declare the sub-fields as 'required' via:

```suggestion
- `featuredImage (required)` - Object with two required properties: `src` and `alt`. The `src` property
```

README.md
```diff
@@ -534,22 +534,22 @@ The `content` section is the body of the post added in Markdown format.
...
   is a path to the featured image which is displayed on the posts list on the`/blog`
-  page. The `alt` property is alternative text for the image. The image should
+  page. The `alt (required)` property is alternative text for the image. The image should
```
bskinn, line 552:
```suggestion
  page. The `alt` property is alternative text for the image. The image should
```

README.md
```diff
@@ -534,20 +534,20 @@ The `content` section is the body of the post added in Markdown format.
...
 - `description` - Description of the blog post. Used inside `<meta name="description" />` tag
-- `published` - Publishing date of the blog post. Used also for sorting posts by
+- `published (required)` - Publishing date of the blog post. Used also for sorting posts by
   date (the format should be `Month d, yyyy` for example `January 1, 2023`)
-- `author` - Unique slug of the author (from Storyblok) usually looks like `jon-doe`.
+- `author (required)` - Unique slug of the author (from Storyblok) usually looks like `jon-doe`.
   Based on this property blog post page will display proper info about the author
   (and their avatar). The author must be present in Storyblok in order for the
   post to build without error.
 - `category` - Array of categories for example `[Machine Learning]`. All
```
bskinn, line 546: @gabalafou -- Do we know if this field is optional, or if it needs to be _present_, but can be an empty list?

Do we _want_ to allow someone not to declare a category?

gabalafou: Honestly no idea. I don't think I've seen any blog post without a category though.

gabalafou: For simplicity, I would just tell the blog author that all fields are required, even if that's not enforced by the machinery.

bskinn, line 546:
```suggestion
- `category` (required) - Array of categories for example `[Machine Learning]`. All
```

README.md
```diff
@@ -534,20 +534,20 @@ The `content` section is the body of the post added in Markdown format.
...
   categories should be the same as in the previously mentioned
   [`categories.json`](./apps/labs/posts/categories.json) file.
   **Important note:** categories are case-sensitive.
-- `featuredImage` - Object with properties: `src` and `alt`. The `src` property
+- `featuredImage (required)` - Object with two required properties: `src` and `alt`. The `src` property
```
bskinn, line 550:
```suggestion
- `featuredImage` (required) - Object with two required properties: `src` and `alt`. The `src` property
```

bskinn (commented):
Adding "(required)" to `category`, and removing code formatting from the other "(required)" annotations

PR #742, opened 2023-06-16
Remove unused dependencies

Surafel-Getachew:
This branch is intended to research on unused npm dependencies and remove them from `package.json`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
         "@tailwindcss/typography": "^0.5.2",
         "axios": "0.27.2",
         "clsx": "^1.2.1",
-        "core-js": "^3.25.1",
         "feed": "^4.2.2",
         "graphql": "^16.3.0",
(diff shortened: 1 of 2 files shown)


Surafel-Getachew:
Approach used to search for unused dependencies:

- Use `npx depcheck` to list out potential unused dependencies.
- check on the library how it could be used, imported and search all the files in the project for that can kind of use.
- Check git history of that line to get context of when it's added, files that are changed during that commit, file that imported that package.
- If we get a use case that can be used leave it as is, else uninstall the package this will remove it from `package.json` and `package-lock.json`.
- Run the app both consulting and labs and check if it cause's any error while running the app.
- Build the app and check if there's any error in the build process.

Packages that are affected/removed by this [commit](https://github.com/Quansight/Quansight-website/pull/742/commits/707670dedb918ed3c8e0c66310586b096f71237e)

- core-js
- regenerator-runtime
- slick-carousel
- swr

Issue #668, opened 2023-02-14
[BUG] - Copy-pasted featured images


gabalafou:
### What site is this for?

Quansight Labs

### Actual behavior

Too many blog posts have copy-pasted the exact same image under `/public/posts/[slug]/`, we should delete the copy-pasted images, hoist them up a level or two, and share them across the blog posts. This will require many image file deletions, and for each deletion, an update to the corresponding post's .md(x) file metadata. For example:

Same images: 
 
1. [apps/labs/public/posts/Spyder4-variable-explorer/blog_feature_org.svg](https://github.com/Quansight/Quansight-website/blob/ad871f8cc9ca71c6a6af7e2455e5cbdb0f624f1d/apps/labs/public/posts/Spyder4-variable-explorer/blog_feature_org.svg)
2. [apps/labs/public/posts/a-second-czi-grant-for-numpy-and-openblas/blog_feature_var1.svg](https://github.com/Quansight/Quansight-website/blob/ad871f8cc9ca71c6a6af7e2455e5cbdb0f624f1d/apps/labs/public/posts/a-second-czi-grant-for-numpy-and-openblas/blog_feature_var1.svg)

Corresponding metadata:

1. [Spyder4-variable-explorer.md line 8](https://github.com/Quansight/Quansight-website/blob/ad871f8cc9ca71c6a6af7e2455e5cbdb0f624f1d/apps/labs/posts/Spyder4-variable-explorer.md?plain=1#L8)
2. [a-second-czi-grant-for-numpy-and-openblas.md line 8](https://github.com/Quansight/Quansight-website/blob/ad871f8cc9ca71c6a6af7e2455e5cbdb0f624f1d/apps/labs/posts/a-second-czi-grant-for-numpy-and-openblas.md?plain=1#L8)

Issue #666, opened 2023-02-14
[BUG] - Too many blog posts have the same featured image (thumbnail)


gabalafou:
### What site is this for?

Quansight Labs

### Actual behavior

![screenshot showing a page with five blog posts that have the same thumbnail](https://user-images.githubusercontent.com/317883/218804068-e3ff3d9c-7175-4b4c-8642-834ad3cf4bb2.png)

### Anything else?

Conversation from Slack:

Gabriel
I don’t want to hold back migrating blog posts for issues like the thumbnail, but I feel like we need to rethink the way we’ve designed the Labs blog, with every blog post requiring a thumbnail—because I feel like it’s leading to pages that look like this, in which every blog post shown on the page uses the same thumbnail image.

trallard
Ditto - think also we had discussed having a greater range of ready to use images.... But maybe having another way to display the posts would be best

PR #751, opened 2023-07-11
Release 2023 07 11

gabalafou:


trallard (approved):

bskinn (approved):
Both blog posts render ok in gallery and individual pages.

Both posts show Labs logo image in preview cards on Twitter and LinkedIn.

The only hitch is stylistic: both posts use the same preview image, so the gallery looks a bit over-heavy with that same image.

image

Not a fatal blocker, so approving. @trallard, do you want to change this before we build to live?

gabalafou:
> The only hitch is stylistic: both posts use the same preview image, so the gallery looks a bit over-heavy with that same image.

This is a longer/bigger issue with the blog gallery. I've raised this before. Do we know if we have an issue tracking this? If not, I'll create one.

bskinn:
#666 is about the philosophy of duplication of the gallery images.

#668 is more logistical: proposing pulling reused images into a common directory, instead of having an independent copy of each image for each blog post that uses it.

gabalafou:
Thanks, @bskinn!

PR #791, opened 2023-10-16
[BLOG] Add numba engine in pandas apply

lithomas1:
## Text styling

## Non-text contents

Discussed with Ralf as part of the numba work for pandas on the NASA Roses Grant.


apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,463 @@
+---
+title: 'Unlocking C-level performance in pandas DataFrame.apply with numba'
+authors: [thomas-li]
+published: October 15, 2023
+description: 'A high-level overview of the new numba engine in DataFrame.apply'
+category: [PyData ecosystem]
+featuredImage:
+  src: /posts/numpy-python-api-cleanup/numpy-python-api-cleanup-featured.png
```
lithomas1, line 8: I should generate an image for this.

@MarcoGorelli Is there any existing pandas image I can re-use?

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,463 @@
...
+```
+
+```python
+# Timing the compilation time
+# To estimate this, we will use the numba engine on a very small DataFrames
+small_df = pd.DataFrame({"a": [1]})
+%time small_df.apply(f, engine="numba", axis=1)
+```
+
+    CPU times: user 312 ms, sys: 6.93 ms, total: 319 ms
```
lithomas1, line 239: This doesn't seem to render properly. (It just renders as plain text)

Is there a markdown tag that renders this not as text or code but as console output?

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,463 @@
...
+It does this by "mocking" the Series object passed in to your function. This is necessary because numba code cannot access the methods on a
+Series like Python code can - each numba equivalent must have either rewrite or wrap the logic behind the original method.
+
+It also opens up the door to parallel `DataFrame.apply`, when operating on the raw values of the DataFrame, which is super exciting as this has been a highly requested feature in pandas over the years.
+
+While many features/methods may be missing from the mocked Series object currently (such is the nature of mocking) and the numba engine in general, it's expected that these will slowly be added in as the numba engine is adopted, and from contributions from users like you!
+
+## Acknoledgements
+
+I'd like to thank [Matthew Roeschke](https://github.com/mroeschke) for reviewing my PRs, and NASA for funding the grant to make this possible.
```
lithomas1, line 463: Is there somewhere else where we acknowledge the NASA grant?

It'd be good to be able to copy that over to make sure I'm not leaving anyone out that was involved in getting the grant.

lithomas1 (commented):
@MarcoGorelli this should be ready for a peer review now

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,463 @@
...
+
+```python
+# Let's take a look at the performance of just a mean operation with numba
+# (both raw mode and parallel on)
+# and just with numpy/pandas
+
+f = lambda x: x.mean()
+%timeit df.apply(f, engine='numba', raw=True, engine_kwargs={'parallel': True})
+%timeit df.mean(axis=1)
+%timeit df.values.mean(axis=1)
```
MarcoGorelli, line 440: is this the same operation though?

```python
In [1]: df = pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})

In [2]: f = lambda x: x.mean()

In [3]: df.apply(f, engine='numba', raw=True, engine_kwargs={'parallel': True})
Out[3]:
a    2.0
b    5.0
dtype: float64

In [4]: df.mean(axis=1)
Out[4]:
0    2.5
1    3.5
2    4.5
dtype: float64
```

MarcoGorelli: I think you want `df.mean(axis=0)` for it to be the same operation?

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,463 @@
...
+%timeit df.apply(f, engine="numba", axis=1, raw=True)
+```
+
+    307 µs ± 44.3 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
+
+We can see here that the difference is several orders of magnitude.
+This is because the unboxing process currently individually unboxes each resultant Series to a Python object
+before running concat.
+
+The solution to this is to do the concatenation inside of numba, so
```
MarcoGorelli, line 307: so...

lithomas1: Thanks!

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,466 @@
...
+We can see here that the difference is several orders of magnitude.
+This is because the unboxing process currently individually unboxes each resultant Series to a Python object
+before concatenating them together to build the result DataFrame.
+
+In future pandas releases, this can be optimized by concatenating the results of apply inside of the numba
+engine (which would only require one final unobxing call for the concatenated DataFrame),
+which should bring the speed of apply on Series in numba on par with the speed of
+apply on the raw numpy arrays.
+
+Despite this, the numba engine is still much faster than the Python engine, as seen below.
```
rgommers, line 312: My impression here is that it helps to spell it out and not let the reader do the math. ~6x speedup for `Series`, and ~6000x for raw numpy arrays!

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,466 @@
...
+%timeit df.values.mean(axis=1)
+```
+
+    178 µs ± 60.9 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
+    1.1 ms ± 99.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
+    32.9 µs ± 2.74 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
+
+Here, we can see that although numba beats the pandas mean function (it's a bit slower than the numpy version because it does some extra work),
+it loses to the numpy mean function, illustrating the point above.
+
```
rgommers, line 452: You could consider if it's worth adding a graph, a la https://github.com/rgommers/explore-array-computing/blob/master/explore_xnd.ipynb? I quite like such log-log plots, they're visually appealing and it makes the scaling with data size (here cols and/or rows) easy to understand.

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,466 @@
...
+All in all, the numba engine will offer a faster way for pandas users to run their apply functions with minimal changes, for functions containing standard pandas operations (think indexing, arithmetic, and using numpy functions on the data) in pandas 2.2 and above.
+
+It does this by "mocking" the Series object passed in to your function. This is necessary because numba code cannot access the methods on a
+Series like Python code can - each numba equivalent must have either rewrite or wrap the logic behind the original method.
+
+It also opens up the door to parallel `DataFrame.apply`, when operating on the raw values of the DataFrame, which is super exciting as this has been a highly requested feature in pandas over the years.
+
+While many features/methods may be missing from the mocked Series object currently (such is the nature of mocking) and the numba engine in general, it's expected that these will slowly be added in as the numba engine is adopted, and from contributions from users like you!
+
+## Acknoledgements
```
rgommers, line 464: typo here

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
rgommers: Either way, one visual would be useful to have. In the presentation you showed me for the Pandas developer meeting you had some Snakeviz profiles, did you consider including one of those?

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,466 @@
...
+  - e.g. addition, subtraction, multiplication, and division
+- Parallel apply
+  - This is only supported when `raw=True`
+  - Like other functions that support numba, you can pass in `parallel: True` inside the `engine_kwargs` dictionary argument to use apply in parallel.
+
+\*: This indicates partial support
+
+Because the numba engine is still very new and experimental, support does not include for now:
+
+- Support for Apache Arrow arrays
```
rgommers, line 65: The bullet lists are visually not great when used too heavily. How about making this a regular paragraph instead?

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,466 @@
...
+
+## How the numba engine works
+
+Currently, in pandas, there are currently two implementations of the numba engine inside of apply, one for `raw=True`, and one for `raw=False` (the default).
+
+If you didn't know already, passing in `raw=True` to `DataFrame.apply` tells pandas to pass in numpy arrays, not pandas Series to the functions you pass into apply. Because, numba itself supports numpy arrays natively, in general, every function runnable with the `Python` engine, should also be runnable with the `numba` engine, as long as you stick to the [supported numpy features](https://numba.readthedocs.io/en/latest/reference/numpysupported.html) inside numba.
+
+When `raw=False` is passed in or no `raw` parameter is passed in, we must "mock" the pandas `Series` object that would normally get passsed in, since numba doesn't natively support pandas DataFrames. This means that while most functions passed into apply with work with the numba engine out of the box with `raw=False`, some functions may require modifications, as the "mocked" DataFrame does not implement the entire pandas API.
```
rgommers, line 42: I think a lot of the complexity of the work came from this mocked Series/DataFrame implementation, right? It would be nice to say a bit more about it. You did some quite interesting work there I believe, however this doesn't really come out in this post yet. So 1-2 paragraphs about some of the key things you had to implement or work around in order to get things to work would help the story I think. Plus pointing at one or a couple of issues/PRs can't hurt.

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
rgommers: In talks perhaps, but no other blog post yet. I suggest to add the following sentence: _"This work was supported by a grant to Pandas, scikit-learn, SciPy and NumPy under the NASA ROSES 2020 program."_

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,466 @@
+---
+title: 'Unlocking C-level performance in pandas DataFrame.apply with numba'
```
rgommers, line 2: Nice title, I like it. Minor: perhaps add a `:` (or a `'`) after `pandas` to connect the two parts of the title? Also, Numba with a capital N (also on line 5).

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
rgommers: I added two suggestions to generate a graph/graphic. Maybe a reworked version of one of those could serve as the header image too.

rgommers (commented):
Thanks @lithomas1! This looks like a great start. I added a few suggestions for content and for making the benchmarks and visual appearance of the post as a whole a bit more polished.

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
lithomas1: Updated.

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
lithomas1: I updated it to code blocks. I'm thinking about using IPython blocks for the rendering (since you can't use magics like timeit in a regular Python repl)

Does that still render fine?

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,466 @@
...
+
+Note that parts a, b, and d contribute to the overhead in running the function. Depending on the function you pass into apply, this overhead can be pretty significant.
+
+We can see the impact of each one of these overheads in our benchmarks below.
+
+### Performance Case Study
+
+Here, we will look at the performance of the numba engine.
+The DataFrame we will operate on will be long and narrow
+(10,000 rows and 3 columns), and we will have a mix of randomly generated integers and floats.
```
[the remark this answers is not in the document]
lithomas1: Updated.

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
lithomas1: Fixed, thanks 🙈

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
lithomas1: Updated.

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
lithomas1: I added a link to the implementation.

There really aren't a lot of issues/PRs taling about this, I basically implemented this in one go in a PR.
(Not sure if there's value linking to my failed attempts here for Groupby.apply. I basically got halfway through on those before I got stuck and had to stop.)

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
lithomas1: Changed it to pandas.DataFrame.apply, that should sound smoother.

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
lithomas1: I'll add a comment to the end, since that also talks about the parallel speed.

This benchmark in particular has also gotten stale, somehow the Python engine for apply caught up (maybe a regression got fixed somewhere), so the numba impl is now a bit slower than Python in the noop case. I'll take a look to see if we can reduce overhead a little more.

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
```diff
@@ -0,0 +1,466 @@
...
+
+We will perform 4 tests here on:
+
+1. Analyzing overhead of the numba engine
+2. Indexing performance of the numba engine
+3. Analyzing performance of the numba engine on normalizing rows of a DataFrame
+4. Comparing raw=True and raw=False, and analyzing the speedups provided by parallel execution with the numba engine when operating on the raw values of the DataFrame.
+
+<details>
+    <summary>DataFrame Creation Code</summary>
```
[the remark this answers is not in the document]
lithomas1: Thanks, updated.

lithomas1:
OK, all comments should be addressed now except for the one about making a plot for perf and the featured image.

I'll make that tomorrow (the overhead for the numba engine is a bit higher than I'd like right now, and I'd like to fix that before generated the perf plots)

apps/labs/posts/unlocking-c-level-performance-in-df-apply.md
lithomas1: The log-log plot makes sense to me. That's probably the most relevant for end users.

I'll probably not include a snakeviz profile, since even though they're helpful to look at, they don't have much meaning for non pandas devs.

lithomas1:
This should be ready for another pass. 

Still not too sure about the hero image/featured image thing. I just used the pandas logo - not too sure if that's appropriate or not.

MarcoGorelli (commented):
good stuff, checked the preview and it looks good. well done!

have just left a really minor comment

rgommers:
Last comment from Marco was addressed too, and this all looks good. So let's get this in!

@bskinn in addition to the hero question, which can be tweaked in a follow-up if we need to, I wanted to ask about the publication date. Is there a release planned for next Tuesday? I'd like to link this post from the year 2 NASA report, which we plan to complete next week.

rgommers:
Very nice work, thanks again Thomas!

bskinn:
@rgommers @lithomas1:

Re the featured/hero image, I think it's fine to use the plain pandas logo from a content perspective, though that's ultimately your call, Ralf, and @trallard's.

I think this _particular_ pandas logo image is not well suited for the hero, though, because it's too low-resolution. When scaled up to hero size, it gets noticeably blurry. So, I think some sort of change here would be a good idea.

If we're going to take that step, though, then I think it would make sense to have a hero created that also has the Numba logo in it, not just pandas. And from _there_, it's a relatively small step to also have a matching featured image created with both logos in it.

Lots of options, just spinning out some ideas.

---

Re the site build, @rgommers, should be no problem to get this post and the CuPy of #802 live early next week. I'll work with @gabalafou and/or @pavithraes on it.

rgommers:
> If we're going to take that step, though, then I think it would make sense to have a hero created that also has the Numba logo in it, not just pandas. And from _there_, it's a relatively small step to also have a matching featured image created with both logos in it.

Better imagery is always nice, however I don't know what that takes in terms of effort/lead-time - so I'd be happy to leave that decision to you.

> Re the site build, @rgommers, should be no problem to get this post and the CuPy of #802 live early next week. I'll work with @gabalafou and/or @pavithraes on it.

thank you!

PR #798, opened 2023-10-31
Release 2023-10-31

bskinn:
Publishing @asmeurer's SymPy documentation blog post for Labs


asmeurer:
Making the header image of the post a screenshot of the docs really doesn't look good https://labs-nlldffix4-quansight.vercel.app/blog/sympy-documentation. It looks like it's part of the page except it actually isn't. I would suggest ditching the header image if that's possible. Or if it isn't, using one of the basic ones from the other posts. 

It also doesn't look great in the listing compared to the other images, but it isn't really as big of a deal there https://labs-nlldffix4-quansight.vercel.app/blog

asmeurer:
If you want an image to use for the blog listing I would suggest just using the SymPy logo 
sympy-500px

Here's a version without the text if that's better 
sympy-notext-500px

On the page itself, like I said, it would be better to have no header image. On my laptop, the header image takes up the whole screen and you can't even see the title of the post without scrolling down (actually, every post seems to have this problem). If that's not actually possible, let's just use one of the stock images.

bskinn:
@asmeurer I noted that, but thought it was an intentional choice so didn't bring it up.  We can still hotfix the images, no problem.

I tossed this together in a few minutes, what do you think of it as the hero?

image

bskinn:
Though, it looks like it partly crops it off at the top, so... hard to tell what it'll look like without actually applying it to the post. I can check locally, though.....

bskinn:
Makes sense to use the SymPy logo for the gallery image, I'll lay that in.

bskinn:
Nah, that "hero" I made is too blurry. This is why they don't pay me to do design.

I don't immediately see any of the heroes from other posts that are a natural fit. Do you have a preference for which one to use, @asmeurer?

(We can always publish with something approximately in the ballpark now, and then ask Irina to make something better and swap it in in a future site build, if desired.)

pavithraes (approved):
Looks good to me, thank you!

PR #799, opened 2023-11-01
[HOTFIX] Release 2023-11-01

bskinn:
Transparent background doesn't work against dark mode for the featured image for @asmeurer's new SymPy post.

Replaced it with a version with white background.


pavithraes (approved):
Thanks for being on top of this, looks good to me!

PR #803, opened 2023-11-08
[BLOG FIX] Switch featured image for Numba polynomials post & edit NumPy post date format

bskinn:
For the Numba post, removing the featured image with the Nebari logo in favor of one with Numba and NumPy logos.

For the NumPy post, switching to a non-zero-padded day in the publish date. Oh, and it looks like my VSCode autoformatted some things in the Markdown.


rgommers (approved):
LGTM, thanks Brian.

PR #807, opened 2023-11-28
[BLOG] Switch to large-field pandas logo for hero

bskinn:
Swapping out the hero image from #791 for a similar version, but with the pandas logo on a larger field.

This (i) makes the entire pandas logo visible and (ii) eliminates blurring from over-scaling of the image.


rgommers (approved):
Thanks Brian!

PR #802, opened 2023-11-02
[BLOG] Add CuPy CZI grant year 1 report blog post

andfoy:
## Text styling

- [x] The blog is written with [plain language](https://www.plainlanguage.gov/guidelines/) (where relevant).
- [x] If there are headers, they use the proper header tags in order to do so (with only one level-one header).
- [x] All links describe where they link to (for example, check the [Quansight labs website](https://labs.quansight.org/)).
- [x] Any kind of styling that the author uses (for example, bold for emphasis) is consistent throughout the blog.

## Non-text contents

- [x] Blog post featured image is in PNG or JPEG format, **not** SVG.
- [x] All content is represented as text (for example, images need alt text and videos need captions or descriptive transcripts).
- [x] If there are emojis, there are not more than three in a row.
- [x] Don't use [flashing gifs or videos](https://www.w3.org/TR/UNDERSTANDING-WCAG20/seizure-does-not-violate.html).
- [x] If it were to be read as plain text, the blog still makes sense and no information is missing.

PR #808, opened 2023-11-29
Release 2023-11-29

bskinn:
Live build for #802 and #791 & post-merge fixes, plus several Dependabot updates.


apps/labs/posts/categories.json
```diff
@@ -5,6 +5,7 @@
   "Community",
   "Developer workflows",
   "Funding",
+  "GPU",
```
gabalafou, line 8: I'm assuming that the addition of this category was previously discussed, reviewed and approved

gabalafou (approved):
The build seems fine. I clicked on the blog posts and the new category. Looks good

apps/labs/posts/categories.json
bskinn: Given that there was an additional `CuPy` category _also_ added, and then removed after subsequent discussion, I'm assuming that yes, the `GPU` category add is good to remain.

rgommers: That one looked like a good idea to me, yes. I didn't see anyone else comment on it.

bskinn: Are there any other posts in the archive that should have `GPU` backfilled onto?

PR #809, opened 2023-11-29
Sync all nx to same version (13.8.5)

gabalafou:
Fixes https://github.com/Quansight/Quansight-website/issues/810.

Yesterday I merged https://github.com/Quansight/Quansight-website/pull/680 and then I wasn't able to merge any of the other Dependabot updates because the `npx nx lint` command fails. 

Digging into this I found that the Nx docs state very explicitly to [Keep Nx Versions in Sync](https://nx.dev/recipes/tips-n-tricks/keep-nx-versions-in-sync). 

Somehow I doubt that Dependabot is smart enough to know this, so I went ahead and changed `dependabot.yml` to tell Dependabot to ignore all of the Nx packages.


package.json
```diff
@@ -60,15 +60,15 @@
...
-    "@nrwl/cli": "13.8.3",
-    "@nrwl/cypress": "13.8.3",
-    "@nrwl/eslint-plugin-nx": "13.9.7",
-    "@nrwl/jest": "15.9.2",
-    "@nrwl/linter": "15.8.1",
-    "@nrwl/react": "13.8.3",
-    "@nrwl/tao": "13.8.3",
-    "@nrwl/web": "14.7.5",
-    "@nrwl/workspace": "13.8.4",
+    "@nrwl/cli": "13.8.5",
```
gabalafou, line 63: This version number, 13.8.5, comes from line 25 above:

    "@nrwl/next": "13.8.5",

bskinn (approved):
Changes make sense.

Going forward, what's the procedure for updating the Nx packages when updates are needed?

Quarterly Slack reminder to check manually?

(From looking at the dependabot.yml docs, it doesn't appear that there's a way to set this up to notify us that important upgrades are needed, without the bot also creating the update PRs?)

.github/dependabot.yml
```diff
@@ -9,3 +9,8 @@ updates:
       prefix-development: chore
       include: scope
     open-pull-requests-limit: 5
+    ignore:
```
gabalafou, line 12: I don't really know what I'm doing in this file, I'm just pattern matching from the [Dependabot docs page](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore)

bskinn: My take from the docs page is that this will make it so DBot treats it as though the `@nx` and `@nrwl` packages don't exist in the project, and will not do anything whatsoever with them... neither report that updates are needed, nor create update PRs.

I think it's a stronger action than what we actually want, but by my read of the docs it doesn't really give us an intermediate option.

bskinn: The [`groups` feature](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) might be workable?

If, say, we insist that all `@nx` and `@nrwl` updates should be grouped into single PRs?

This working correctly would rely on the assumption that dependabot would 'notice' **all** of the relevant updates all at once. I could see this still glitching if the grouped update PR only included a subset of the `@nx`/`@nrwl` packages that are specified.

.... that said, we now know about having to keep all of them in sync (which we should document somewhere ... in README?), and this `groups` feature _should_ at least cluster all of these version bumps into a single PR...?

gabalafou: I was just about to write and say thanks for asking about how we're going to get updates for `nx` because then I took a second look and also noticed the groups feature. Update incoming

PR #814, opened 2023-11-29
Tell Dependabot to ignore major updates to Nx

gabalafou:


bskinn (approved):
Makes sense.

This still leaves us needing a separate notification mechanism for when `major` Nx version migrations are needed, though.

gabalafou:
I think we're fine without getting notifications for major Nx versions. At the core of this repo are: Node, React, Next, and Nx. To me, getting notifications for those falls under the "nice to have" label. 

But we could try adding a new group "nxMajor" for the major updates if the config file allows us to add a specific commit message prefix for that group. Something like: "DO NOT MERGE (major Nx update require manual update, see Nx docs)"

bskinn:
> I think we're fine without getting notifications for major Nx versions. At the core of this repo are: Node, React, Next, and Nx. To me, getting notifications for those falls under the "nice to have" label.

Agreed.

> But we could try adding a new group "nxMajor" for the major updates if the config file allows us to add a specific commit message prefix for that group. Something like: "DO NOT MERGE (major Nx update require manual update, see Nx docs)"

By my read, this isn't possible -- the commit messages &c appear only to be configurable per-package-registry. The closest we might be able to come is naming the group something like `nxMajorDoNotMerge`, in case the group name is included in the PR title or something.

bskinn:
It appears that DBot _does_ put the dependency group name into the PR title:

image

PR #816, opened 2023-11-30
Upgrade to Nx 15.9.1

gabalafou:
This PR can somewhat be thought of as a re-opening of closed PR #723.

The process to create this PR looked something like:

```
git checkout develop
git pull
git checkout -b 15-9
nx migrate 15.9.1
git commit
npm install --force
git commit
# make some changes to package.json
git commit
npm install
git commit
# add .nx/cache to .gitignore
nx migrate --run-migrations --create-commits
```

Then I cherry picked a bunch of commits from #723.


gabalafou:
Note the following commands run without errors locally:

```
nx serve labs
nx lint labs
nx test labs
``` 

The same commands except `test` also worked for consulting. `nx test consulting` produces an error about `fetch` not being available globally. The error is thrown by the Apollo client constructor. It's puzzling to me because [`fetch` has been a global in Node.js since v18](https://nodejs.org/docs/latest-v18.x/api/globals.html#fetch).

Nonetheless, I get the same error on `develop` when I run `nx test consulting` so it shouldn't hold back this PR.

PR #823, opened 2023-11-30
[MAINT] Raise limit for warning about page size

bskinn:
This PR would raise the limit for page size that triggers a warning during build.

The only effect here is to quiet the build log output. I don't think there's much we can do about the page sizes.

One reason not to do this is because the setting is still `experimental`, and may cause problems with builds later on.

I'm +0.5 to making this change, but would understand a decision not to.

### DIFF
--- a/apps/labs/next.config.js
+++ b/apps/labs/next.config.js
@@ -13,6 +13,10 @@ const nextConfig = {
   images: {
     domains: ['a.storyblok.com'],
   },
+  experimental: {
+    // Only warn about built pages at 1MB and above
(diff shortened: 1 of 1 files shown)


gabalafou:
I've always assumed that something could be done about some of these large page builds but I've never looked into it. If you say it's unavoidable, then sure I'm +1

bskinn:
:wave: @gabalafou @pavithraes Closing this to ~relinquish my ownership of the contribution. If it's still of interest, please recreate as a new PR.

PR #842, opened 2024-04-12
Update Python3 statement url.

Carreau:
See https://github.com/python3statement/python3statement.github.io/issues/292

The devs will not renew the domain forever – it is redirecting for now, but will not do that forever.


rgommers (approved):
Thanks @Carreau!

Carreau:
Thanks !

PR #845, opened 2024-05-17
Release 2024-05-17

gabalafou:


rgommers (approved):
This all looks fine to me, so go ahead and merge if you're happy @gabalafou.

We should really revisit this workflow and get rid of the two-stage deploy. We want PR merges to show up live straight away. That avoids having to do this dance. I'm not sure how easy that is - I'd assume not very. Could you please have a look at that?

rgommers:
@gabalafou I'm also reviewing blog posts right now (one merged, one to go). It'd be great to take those along here while you're at it.

gabalafou:
> @gabalafou I'm also reviewing blog posts right now (one merged, one to go). It'd be great to take those along here while you're at it.

Sounds good, just ping me on Slack when ready

As for getting rid of the two-stage deploy process, here are some thoughts:

- Configure `main` to be the default branch so that blog authors branch off main when working on a blog post
- If a blog post PR is off main, then when it gets merged, it will be deployed straight away (no more two-stage process)
- Leave the `develop` branch in place for riskier changes to the repo that should go through the two-stage deploy process
- Occasionally we'll need to sync (rebase?) the `develop` branch with `main`. Seems to me that `main` should be the source of truth and `develop` should mirror `main`. (The alternative would be merge commits from main in the develop branch but that sounds messy in the long run.)
- Update the docs to reflect the above

What do you think? Make sense? Am I missing anything obvious?

rgommers:
That sounds like a great plan. @trallard do you agree?

PR #872, opened 2024-08-08
[Blog] pandas Resample

MarcoGorelli:
## Text styling

## Non-text contents


apps/labs/posts/pandas-resample.md
```diff
@@ -0,0 +1,133 @@
+---
+title: "Why you probably shouldn't be using `df.resample('M')`"
+authors: [marco-gorelli]
+published: June 17, 2024
+description: 'Learn what you should be using instead'
+category: [PyData ecosystem]
+featuredImage:
+  src: /posts/pandas-resample/pandas-resample.png
+  alt: 'Image of Drake not looking impressed at `resample("M")`, and preferring `resample("MS")`'
```
MarcoGorelli, line 9: image

apps/labs/posts/pandas-resample.md
```diff
@@ -0,0 +1,133 @@
...
+df.resample('MS', on='date')['price'].mean()
+```
+
+Let's learn why!
+
+## What happens when we resample by 'M' (or 'ME')
+
+Before we begin, let's talk about `'M'`. You'd be forgiven for thinking that it stands for "month".
+But actually, looking at the [pandas docs](https://pandas.pydata.org/pandas-docs/version/2.1/user_guide/timeseries.html#offset-aliases), we see that it stands for "month end"! Indeed in pandas 2.1 it was deprecated and
+renamed to `'ME'` to make its meaning clearer.
```
dharhas, line 44: You mention that `ME` means `Month End` here but I don't think you define what `MS` stands for anywhere.

apps/labs/posts/pandas-resample.md
```diff
@@ -0,0 +1,133 @@
...
+
+- Label `'2020-01-31'` corresponds to the bucket `('2019-12-31 23:59:59.999999999', '2020-01-31 23:59:59.999999999']`
+- Label `'2020-02-29'` corresponds to the bucket `('2020-01-31 23:59:59.999999999', '2020-02-29 23:59:59.999999999']`
+- ...
+
+This looks a bit complicated, and probably not at all what users expect. Ask a crowd of data scientists which bucket
+the data point corresponding to `'2020-01-31 01:00:00'` would land in and, unless they've read this article, you'll
+probably get mixed answers.
+
+## What happens when we resample by 'MS'
```
dharhas, line 79: This might be a good place to say `MS` is `Month Start`

MarcoGorelli: done, thanks

pavithraes (approved):
Reads great, thank you! 🌻

PR #873, opened 2024-08-23
Make blog title h1 not h2

gabalafou:
The blog title displayed on the page should be wrapped in an `<h1>` not an `<h2>`.

Note: yes, this means that some blog posts will have two or more h1's. This will occur for blog posts in which the author used one or more level-1 markdown headings. Level-1 headings are the ones with a single hash (#) sign at the beginning of the line. But those are in the minority, and this fixes headings for most of the Labs blog posts. Plus for the blog posts in the minority, the headings were already broken, out of order. They would start with an h2, followed by an h1. Now it starts with an h1 followed by an h1. Not ideal but arguably less broken than before. A follow up PR (#874) will attempt to fix those blog posts.


pavithraes (approved):
LGTM, thanks! 🌻

PR #894, opened 2024-12-20
[BLOG] Arrow PyCapsule Interface and Narwhals for universal dataframe support

MarcoGorelli:
## Text styling

## Non-text contents


MarcoGorelli:
cc @kylebarron @willayd Here's a blog post I've put together on the PyCapsule Interface and Narwhals - just pinging you in case you were interested in reading about the treatment of the former, and whether I've misrepresented anything (if not, no worries at all! Feel free to unsubscribe)

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+Summing non-null values isn't rocket science, but how can we do this in a way that all
+input libraries are supported without any of them being required?
+
+## Low-level solution: PyCapsule Interface in Rust via PyO3
+
+An example of a Rust solution via Pyo3 can be found at [pycapsule-demo/src/lib.rs](https://github.com/MarcoGorelli/pycapsule-demo/blob/6aad64be26ebbfc8526f26695544bfc6436e3266/src/lib.rs#L9-L56).
+
+The technical details are beyond the scope of this post, but the summary is:
+
+- We accept any object which implements the
```
WillAyd, line 53: Not sure how technical you want to be and I realize this post is aimed at dataframe libraries, but there are other methods that can be implemented that still fall under the PyCapsule Interface

 * __arrow_c_array__
 * __arrow_c_schema__
 * __arrow_c_stream__
 * __arrow_c_device_array__
 * __arrow_device_array_stream__

Stream may be the most fitting for exchanging dataframes, but the other ones have use outside of 2-d structures. May want to be careful in saying that the presence of this method exclusively means you are implementing the PyCapsule interface

WillAyd (commented):
looks great

.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
   "editor.formatOnSave": true,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
-    "source.fixAll.format": true
+    "source.fixAll.eslint": "explicit",
```
pavithraes, line 6: To confirm, is this an intentional update?

pavithraes (approved):
Thanks! Everything looks good, and it's an interesting read!

.vscode/settings.json
MarcoGorelli: 😳 - reverted, thanks!

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+  6
+  >>> agnostic_sum_i64_column(rel, column_name="a")
+  6
+  ```
+
+Like magic, our function works agnostically for any of these dataframes,
+without us having to write any specialised code to handle the subtle differences
+between them!
+
+> **_NOTE:_** How many times can `agnostic_sum_i64_column` be called? Strictly
```
WillAyd, line 94: Depending on your target audience, you might be able to remove this note. I love the thoroughness, but it is _maybe_ too into the weeds

I also am not sure I agree with the content - each invocation to the dataframe libraries is going to yield a new Python capsule. In the non-duckdb cases, those capsules provide a non-owning view of the data; the data has a separate lifetime managed by a different object. I think in the duckdb case, they are tying the lifetime of the array to the invocation of the first capsule, but that's more of a technical detail of that library than anything related to the specification itself

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+## Narwhals vs PyCapsule Interface: when to use one over the other, and when to use them together
+
+If the Narwhals API is extensive enough for your use-case, then this is arguably
+a simpler and easier solution than writing your own Rust function. On the other hand,
+if you write a Rust function using the PyCapsule Interface, then you have complete
+control over the data. So, when should you use which?
+
+Let's cover some scenarios:
+
+- If you want your dataframe logic to stay completely lazy where possible: use Narwhals.
```
WillAyd, line 158: This comment isn't quite true either - the PyCapsule Interface says nothing about lifetimes. In the dataframe case, it yields a pointer to a stream of arrays. It is up to the consumer when they want to actually start iterating over the stream

Perhaps the main consumers of the PyCapsule interface choose to do so eagerly, but that's an implementation detail of the libraries themselves, not the PyCapsule Interface

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+  The PyCapsule Interface requires you to materialise the data into memory immediately,
+  whereas Narwhals has both eager and lazy APIs.
+- If you want complete control over your data: use the
+  PyCapsule Interface. If you have the necessary Rust / C skills, there should be no limit
+  to how complex and bespoke you make your data processing.
+- If you want to keep your library logic to pure-Python and without heavy dependencies so
```
WillAyd, line 164: Just a suggestion on ordering - I think most users will be totally fine with narwhals, especially if your target audience is Python developers. So might be worth putting narwhals first on the list

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+  it's easy to maintain and install: use Narwhals. Packaging a pure-Python project is very
+  easy, especially compared with if you need to get Rust or C in there.
+- If you want to do part of your processing in Python, and part of it in Rust - **use both**!
```
WillAyd, line 167: If you cared to cater to non-Rust developers, you might also want to mention nanoarrow. I've also done a youtube video on building Python extensions using C++ - https://www.youtube.com/watch?v=EhUnmXPjTy8&list=PLyPvk-ejh-138onAgM7sVcpzu1HLp-Uiv

Not trying to promote my own content here - just offering in case you wanted to expand the scope of your blog post (which may or may not be worth it)

WillAyd (commented):
Again minor comments and suggestions, but feel free to ignore. Thanks for writing this - glad to see this get more promotion

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+  6
+  ```
+
+Like magic, our function works agnostically for any of these dataframes,
+without us having to write any specialised code to handle the subtle differences
+between them!
+
+> **_NOTE:_** How many times can `agnostic_sum_i64_column` be called? Strictly
+> speaking, the PyCapsule Interface only guarantees that it can be called once for
+> any given input. In practice, however, all implementations that we're aware of
```
kylebarron, line 96: FWIW arro3 IO functions like [`read_parquet`](https://kylebarron.dev/arro3/latest/api/io/#arro3.io.read_parquet) I think will provision a single stream instance. So the first time you use the stream pointer from `read_parquet` it would consume the stream I think and you wouldn't be able to use it again.

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+    lf = nw.from_native(df_native).lazy()
+    schema = lf.collect_schema()
+    if column_name not in schema:
+        msg = f"Column '{column_name}' not found, available columns are: {schema.names()}."
+        raise ValueError(msg)
+    if (dtype := schema[column_name]) != nw.Int64:
+        msg = f"Column '{column_name}' is of type {dtype}, expected Int64"
+        raise TypeError(msg)
+    df = lf.collect()
+    return df[column_name].sum()
```
kylebarron, line 131: What backend do you use to actually compute the sum?

apps/labs/posts/narwhals-pycapsule.md
kylebarron: To be fair, it is possible to write lazy logic in Rust/C++ too. Many of the compute functions exported by `arro3` are lazy. See the overloads for e.g. [`cast`](https://kylebarron.dev/arro3/latest/api/compute/#arro3.compute.cast). If you pass in something with a `__arrow_c_stream__` dunder, the stream will **not** be materialized automatically. Rather, it will return an `ArrayReader`, so that it's fully lazy. Only when the output `ArrayReader` itself is iterated over will the `cast` calls actually happen.

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+
+If the Narwhals API is extensive enough for your use-case, then this is arguably
+a simpler and easier solution than writing your own Rust function. On the other hand,
+if you write a Rust function using the PyCapsule Interface, then you have complete
+control over the data. So, when should you use which?
+
+Let's cover some scenarios:
+
+- If you want your dataframe logic to stay completely lazy where possible: use Narwhals.
+  The PyCapsule Interface requires you to materialise the data into memory immediately,
```
kylebarron, line 159: If anything, I think it's the opposite; the Arrow Stream interface encourages laziness by being defined as a stream. It's only common in-memory implementations that fully materialize data.

apps/labs/posts/narwhals-pycapsule.md
```diff
@@ -0,0 +1,199 @@
...
+  [import](https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.from_arrow)
+  and [export](https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.__arrow_c_stream__).
+
+## What about Polars Plugins?
+
+We wrote some Rust code earlier to express custom logic. You may have come across another way
+to do this in Polars: [Expressions Plugins](https://marcogorelli.github.io/polars-plugins-tutorial/).
+How does that differ from writing custom code with the PyCapsule Interface?
+
+- Pros: Polars Plugins slot in seamlessly with Polars' lazy execution. This can enable massive
```
kylebarron, line 179: FWIW it would be nice to have a generic way to use UDFs defined in terms of the pycapsule interface with Polars. It wouldn't be that hard to write the glue code

kylebarron (approved):
Overall looks good! I've had a long-stalled pycapsule post on my back burner too

kylebarron:
Also in case anyone else is curious, I simplified the rust example: https://github.com/MarcoGorelli/pycapsule-demo/pull/1

apps/labs/posts/narwhals-pycapsule.md
MarcoGorelli: sounds good, happy to link it, just not sure how to phrase it - any suggestions?

apps/labs/posts/narwhals-pycapsule.md
MarcoGorelli: The native one (e.g. Polars in => polars compute, PyArrow in => PyArrow compute, pandas in => pandas compute, cuDF in cuDF compute)

WillAyd:
> tried to simplify the part about how many times you can call `__arrow_c_stream__` per object, but have kept it in. I think it's important to mention, as it is something that users run into

I think this mixes up a duckdb bug or implementation detail with the PyArrow specification. Are you referring directly to this line from the spec?

> Just like in the C Data Interface, the PyCapsule objects defined here can only be consumed once.

That does not mean that you can only call `__arrow_c_stream__` once per object, just that the capsule returned from that can only be used once (much like exhausting an iterator). However, you can create multiple capsules off of the same producer, given the producer does not transfer ownership of the Arrow data to any consumer

MarcoGorelli:
Sure, thanks, I've shortened the note to just observe the behaviour without getting too much into the details. Based on https://github.com/duckdb/duckdb/discussions/15536#discussioncomment-11726340, neither they nor Kyle consider it a bug. I can't really comment one way of another, it's just something I'd like users to be aware of

apps/labs/posts/narwhals-pycapsule.md
kylebarron: I don't think it would be that hard to do. Would you like to pair on it?

MarcoGorelli: would love to!

MarcoGorelli:
Thanks all for reviews and comments! Doesn't look like there's been further issues raised, so I'll go ahead and ship this one - if there are any other comments, please do feel free to comment, we can easily and speedily correct them!

PR #895, opened 2025-01-08
FIX - make team member image container inline-flex

gabalafou:
@agriyakhetarpal reported in Slack:

> Hi! I noticed a minor issue on the Labs team page (please see attached screenshot), where there's an empty pastel lavender coloured bar on the right of everyone's photos for "page width (px)" ∀ [1185, 1280). I could propose a fix for the CSS myself, but AFAIK, we don't have this page in specific open-sourced.
> <image src="https://github.com/user-attachments/assets/6d0f5812-c12c-42ed-8def-3444d4f7fd28" alt="" width="400"/>

This small change seems to do the trick.


agriyakhetarpal (approved):
Thanks for the quick fix, @gabalafou!

pavithraes (approved):
LGTM, thanks! 🌻

PR #897, opened 2025-01-17
[BLOG] PEP 517 build system popularity

mgorny:
The first complete draft, with text and tabular data. Originally, I wanted to include some pie charts too, but I'm not sure anymore — the data doesn't look particularly "chartable" (say, setuptools ending up taking 3/4 of the pie). Table 5 ended up pretty big.

In general, I'm not very good all the visual stuff, so I'd appreciate some hints. I've put a hero image but not "featured" image yet. However, I'm slightly confused here — Notion seems to give specific dimensions for these images, but blog posts don't seem to follow them, so I don't know if I should resize/crop them or not.

## Text styling

## Non-text contents


mgorny:
Ok, I think this is ready for a first round of review.

rgommers (commented):
Thanks Michal. I had a first read. Looks interesting. I can't read the tables and those are quite important for the overall story here, so I'll have another look once the preview works.

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,558 @@
...
+while using another PEP 517 build backend. Most likely, these packages
+used a backend that did not provide direct support for building C
+extensions, and used setuptools to provide that function. In fact,
+some build backends officially support that.
+
+Secondly, over half of the packages depending on `setuptools`
+additionally depended on the `wheel` package. Sometimes `wheel` is actually
+used by the `setup.py` script, but most often this is copied from
+[a historical mistake in setuptools documentation that listed `wheel`
+dependency unnecessarily](https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a).
```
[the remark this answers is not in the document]
mgorny: I can't think of a reason why. The PEP 517 backend did add it automatically from day one, so the only reason it wouldn't work would be if some tool didn't implement PEP 517 correctly.

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,558 @@
...
+Many backends were developed as part of some packaging tool. In my experience,
+many people choose `hatchling`, because they use Hatch. They choose
+`flit_core` when they use Flit, `poetry-core` when they use Poetry,
+`pdm-backend` when they use PDM. Some of them think they cannot use
+another backend with their chosen tool. Some people probably choose
+Hatchling, because it is the default option in [‘Choosing a build
+backend’ part of the Python Packaging User
+Guide](https://packaging.python.org/en/latest/tutorials/packaging-projects/#choosing-a-build-backend).
+
+Over half of the packages using setuptools still rely on `setup.py`.
```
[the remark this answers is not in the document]
mgorny: Hmm, I messed up that sentence while rewriting. Will fix in a minute — I meant to say over half (64% to be precise) declare metadata only in `setup.py`.

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,558 @@
...
+Hatchling (6.5% packages) and Flit (3.8%). Build systems other than the four
+already mentioned amount for 2.6% of packages. They include both generic Python build
+systems (such as `pdm-backend`), as well as more specialized tools such as:
+
+- Maturin — used to build Rust packages (1.1%)
+
+- scikit-build-core — used to integrate with CMake build system (0.40%)
+
+- mesonpy — used to integrate with Meson build system (0.22%)
+
```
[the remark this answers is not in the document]
mgorny: Hmm, can't think of a really clean way of distinguishing them, that wouldn't give a large number of false negatives/positives.

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,558 @@
...
+  <tr style={{ background: '#efe' }}><th colspan='2'>pdm</th><th></th></tr>
+  <tr style={{ background: '#efe' }}><td></td><td>`pdm.backend`</td><td align='right'>37</td></tr>
+  <tr style={{ background: '#efe' }}><td></td><td>`pdm.pep517.api`</td><td align='right'>4</td></tr>
+  <tr style={{ background: '#efe' }}><td></td><td>`pdm.backend.intree`</td><td align='right'>1</td></tr>
+  <tr style={{ background: '#fee' }}><th colspan='2'>scikit-build-core</th><th></th></tr>
+  <tr style={{ background: '#fee' }}><td></td><td>`scikit_build_core.build`</td><td align='right'>28</td></tr>
+  <tr style={{ background: '#fee' }}><td></td><td>(custom)</td><td align='right'>2</td></tr>
+</table>
+
+</div>
```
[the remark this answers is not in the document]
mgorny: Hmm, I was so focused on trying to figure out how make this into a nice pie chart, I didn't think of using a bar chart. But yeah, that could work. I suppose I could also try using a stacked bar chart to split families into specific backends.

apps/labs/posts/pep-517-build-system-popularity.md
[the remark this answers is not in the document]
rgommers: Ah sure, I meant `python setup.py bdist_wheel` needed it, and hence people just carried it over to `pyproject.toml` without much thinking.

apps/labs/posts/pep-517-build-system-popularity.md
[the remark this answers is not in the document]
mgorny: Hmm, I don't think bar chart is going to work here. With [the numbers we have](https://labs-jhtqsa5rk-quansight.vercel.app/blog/pep-517-build-system-popularity#table-1), linear plot would be completely unreadable, and I think logarithmic scale would be confusing. I'm going to try with a pie chart instead.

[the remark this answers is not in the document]
mgorny: Actually, I've messed up the numbers somehow. Gotta fix that first.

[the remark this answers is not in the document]
mgorny: Scratch that, my notebook had a typo and I've been summing wrong. The numbers are okay.

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,584 @@
+---
+title: 'PEP 517 build system popularity'
+published: January 24, 2025
+authors: [michal-gorny]
+description: 'A closer look at non-elementary group-by aggregations'
```
MarcoGorelli, line 5: I recognise this sentence from somewhere 😉

mgorny: Haha, sorry, forgot to update it :-).

mgorny:
> the last table (P/H) is very long and hard to read - perhaps that one can be broken up sensibly (e.g., separate backend plugins from regular build dependencies, and maybe versioning-related tools is large enough to be a separate category?).

Yeah, I don't like it either. I need to try a bit and see how it works out.

mgorny:
Ok, I've split the tables a bit — but I'm not sure if that really made things much better. Perhaps I should just ditch having per-backend counts in the subsequent tables, as they really matter only for setuptools.

mgorny:
I've removed the per-backend counts from all but the first table, and I think they look much better now. One more thing I could do (but no sure if there's a point in actually doing it) would be transposing table 5, so it's good only "setuptools" and "wheel" columns, and rows for backends.

mgorny:
Just for posterity: I had a crazy idea, but it doesn't look as good as I wished it would:

badidea

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,615 @@
...
+  <caption>Table 1. Cumulative backend use counts</caption>
+  <tr><th>Family or backend</th><th>Count</th></tr>
+  <tr style={{ background: '#fef' }}><td style={{ height: '4em' }}>setuptools</td><td align='right'>5854</td></tr>
+  <tr style={{ background: '#eff' }}><td style={{ height: '4em' }}>poetry</td><td align='right'>625</td></tr>
+  <tr style={{ background: '#ffe' }}><td style={{ height: '4em' }}>hatchling</td><td align='right'>480</td></tr>
+  <tr style={{ background: '#eef' }}><td style={{ height: '4em' }}>flit</td><td align='right'>285</td></tr>
+  <tr><td style={{ height: '3em' }}>`maturin`</td><td align='right'>85</td></tr>
+  <tr style={{ background: '#efe' }}><td style={{ height: '4em' }}>pdm</td><td align='right'>42</td></tr>
+  <tr style={{ background: '#fee' }}><td style={{ height: '4em' }}>scikit-build-core</td><td align='right'>30</td></tr>
+  <tr><td style={{ height: '3em' }}>`mesonpy`</td><td align='right'>16</td></tr>
```
rgommers, line 107: minor: could you rename to `meson-python`

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,615 @@
...
+
+<table id='table-5' style={{margin: 'auto', width: 'auto'}}>
+  <caption>Table 5. Setuptools and wheel dependencies</caption>
+  <tr><th>Build backend</th><th>setuptools</th><th>wheel</th></tr>
+  <tr style={{ background: '#fef' }}><td>setuptools</td><td align='right'>1891</td><td align='right'>1071</td></tr>
+  <tr style={{ background: '#eff' }}><td>poetry</td><td align='right'>24</td><td align='right'>10</td></tr>
+  <tr style={{ background: '#ffe' }}><td>hatchling</td><td align='right'>6</td><td align='right'>0</td></tr>
+  <tr style={{ background: '#efe' }}><td>pdm</td><td align='right'>1</td><td align='right'>0</td></tr>
+  <tr style={{ background: '#fee' }}><td>scikit-build-core</td><td align='right'>1</td><td align='right'>0</td></tr>
+  <tr><td>`mesonpy`</td><td align='right'>1</td><td align='right'>1</td></tr>
```
rgommers, line 395: minor: could you rename to `meson-python`

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,615 @@
...
+- 90 packages missing some required files (such as README or requirement
+  files specified in `setup.py`)
+
+- 64 missing Python dependencies in `build-system.requires` key
+
+The remaining packages were either failing due to incorrect metadata,
+use of removed setuptools API, code incompatible with Python 3.11
+or expecting being built from a git checkout.
+
+<table id='table-5' style={{margin: 'auto', width: 'auto'}}>
```
rgommers, line 387: There's a mix because project/package names (in black font) and import names (in pink font), is this on purpose? It looks a little off to me, especially for names which match (e.g., `maturin` could easily be black).

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,615 @@
...
+
+Setuptools was used as a build system for almost 79%
+of the tested packages. Seven out of ten packages using setuptools do not
+declare a build backend in `pyproject.toml` — they rely on the tools
+running `setup.py` when no backend is declared.
+
+The three next top ranking build system families are Poetry (8.4% packages),
+Hatchling (6.5% packages) and Flit (3.8%). Build systems other than the four
+already mentioned amount for 2.6% of packages. They include both generic Python build
+systems (such as `pdm-backend`), as well as more specialized tools such as:
```
rgommers, line 181: I'm not sure the split is generic vs. specialized - pdm and the backends other than setuptools in the sentence before are all for pure Python code which Maturin/scikit-build-core/meson-python are the main options (next to setuptools) for packages with compiled code. Perhaps better to phrase it like that?

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,615 @@
...
+easy to make mistakes such as:
+
+```python
+install_requires = []
+# WRONG!
+if sys.version_info < (3, 11):
+    install_requires.append("exceptiongroup")
+```
+
+With the above snippet, if a wheel is built with Python 3.10, it will
```
rgommers, line 239: At first I thought you meant sdist, not wheel, here. However, now I think it is wheel but `py3-none-any` wheels specifically. Correct? It reads a bit confusing to me, because I was thinking about `cp31x` wheels.

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,615 @@
...
+    and two smaller circles representing setup.cfg and pyproject.toml.
+    The setup.cfg circle is mostly overlapping with setup.py, while
+    the pyproject.toml circle is closer to half-overlapped."
+  />
+
+  <figcaption>Figure 3. Use of setuptools configuration formats combinations</figcaption>
+</figure>
+
+`setup.py` still remains the most popular of the configuration formats.
+It is used by 89% of the analyzed packages, with 64% not using any other
```
rgommers, line 303:
```suggestion
It is used by 89% of the analyzed setuptools-using packages, with 64% not using any other
```

or something similar. just to avoid the "that's larger than all packages using setuptools together" response

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,615 @@
...
+  files specified in `setup.py`)
+
+- 64 missing Python dependencies in `build-system.requires` key
+
+The remaining packages were either failing due to incorrect metadata,
+use of removed setuptools API, code incompatible with Python 3.11
+or expecting being built from a git checkout.
+
+<table id='table-5' style={{margin: 'auto', width: 'auto'}}>
+  <caption>Table 5. Setuptools and wheel dependencies</caption>
```
rgommers, line 388: No change to the blog post needed, but: I'd be curious to learn which packages were the `1` for scikit-build-core and for meson-python.

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,615 @@
...
+`setuptools-rust`, and further 12 used `scikit-build` (also note that 30
+projects used `scikit-build-core`, per table 1).
+
+32 packages declared a dependency on `cmake`, and 16 on `ninja`. These
+PyPI packages provide precompiled executables for systems where system
+tools are not available. Some projects add these dependencies only
+if they cannot find the required tool. Since these tools were present
+on my system, the numbers here represent packages adding
+the dependencies unconditionally. This indicates that they will use
+a local copy instead of the system tools.
```
rgommers, line 502: "local copy" -> "wheel-installed version" or something like that, for clarity?

rgommers (commented):
This looks very close to ready, and I enjoyed reading it top to bottom again. A few localized things to tweak, nothing major.

mgorny:
> @mgorny I see you switched from `.png` to `.svg`, and the guidelines in the PR template disagree (I can't remember if the reason for it had to do with accessibility or portability):
> 
> > Blog post featured image is in PNG or JPEG format, not SVG.
> 
> could you revert that change?

As discussed on Slack, disregarding that as the guidelines apply to the "featured" image, and that one's JPEG.

apps/labs/posts/pep-517-build-system-popularity.md
mgorny: I can, though I'm using the backend Python package name here, which is `mesonpy` (it's taken straight from `pyproject.toml`).

rgommers: The Python package name is `meson-python`; the import name is `mesonpy`. You're also using `scikit-build-core` for example (with dashes), which isn't the import name but the package name. There's a bit of inconsistency here. I think package names are clearer.

apps/labs/posts/pep-517-build-system-popularity.md
mgorny: Yes, "kinda". The idea is that multi-backend families (i.e. those in sans-serif font) are further extended in table 2, whereas single-backend options are only in table 1. I suppose I could just replace most of them with proper names — though for the few last options this could be confusing since pbr and jupyter-packaging can be used both via PEP 517 backends and as setuptools plugin, and only the former are counted here.

Perhaps I could use proper names down to "(custom)" (or even sphinx-theme-builder), and import names below? I suppose that would look better, and remain clear for the multi-component backends.

apps/labs/posts/pep-517-build-system-popularity.md
mgorny: Yes, I meant pure Python wheels. I wish the word "universal" wouldn't be used already for py2/py3 wheel, but perhaps it'd not be too technical to say `py3-*` wheel here.

apps/labs/posts/pep-517-build-system-popularity.md
mgorny: scikit-image for mesonpy, soxr for scikit-build-core. Both seem unused.

Technically, [scikit-build-core has a setuptools variant](https://scikit-build-core.readthedocs.io/en/latest/plugins/setuptools.html) but I didn't see that used in the wild (yet).

apps/labs/posts/pep-517-build-system-popularity.md
rgommers: That sounds like a good solution to me.

apps/labs/posts/pep-517-build-system-popularity.md
rgommers: Maybe `universal (py3-*)`?

Or otherwise "a wheel from pure Python code"?

mgorny:
Addressed all remarks.

rgommers (approved):
All right, let's ship it! Nice post, thanks Michal.

apps/labs/posts/pep-517-build-system-popularity.md
```diff
@@ -0,0 +1,616 @@
+---
+title: 'PEP 517 build system popularity'
+published: January 24, 2025
```
rgommers, line 3:
```suggestion
published: February 11, 2025
```

PR #905, opened 2025-02-11
PEP 517 build system popularity: fixes to figure 3 readability

mgorny:
As suggested by @dharhas, move `setup.cfg` group label to the right. While at it, also color-code the matching tables to make them look nicer and suggest relationship with the diagram.


rgommers (approved):
LGTM, in it goes. Thanks @mgorny

PR #913, opened 2025-03-20
Small fixes for 2024 report blog post

gabalafou:
This fixes a couple small issues with the blog post for the 2024 annual report:

- In React, the attribute should be `className` not `class`.
- Our framework automatically linkifies emails, and since we were manually linking the email, this meant that it was linked twice


gabalafou:
This is not urgent since the correct HTML is output, but it causes an error to be printed to the logs.

pavithraes (approved):
Thanks! Since it's not urgent, I think we can wait to merge this to include the homepage update on Storyblok (which should be done later today).

PR #914, opened 2025-03-20
Linkify and lock size of infographic (2024 annual report)

gabalafou:
This PR fixes a two different issues with the infographic

1. Links in graphic not clickable
2. Infographic too small on mobile (see screenshot below)

To fix both of these issues, this PR:

1. Adds an image map
2. Locks the width and height of the image (also necessary for the image map)

Yes, this means that the user has to scroll if the screen is too narrow (less than about 900px). However, some considerations:

1. On mobile, the user already has to pinch zoom and then scroll horizontally for the infographic to be readable
2. On tablets, the user can reorient to landscape instead of portrait and then the infographic will likely fit
3. Most desktop screens are greater than 900px, so there shouldn't be any problem on desktop.

One thing I did not do was actually resize the PNG file itself. I figured that if somebody wanted to download the image and/or print it, it would be better if they have the full resolution.

mobile screenshot

### DIFF
--- a/apps/labs/posts/labs-annual-report-2024.md
+++ b/apps/labs/posts/labs-annual-report-2024.md
@@ -142,7 +142,7 @@ The following graphic presents notable highlights from the report.
 
   <section>
     <h3>Contact Us</h3>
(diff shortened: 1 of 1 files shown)


pavithraes (approved):
Thanks, LGTM! 🌻

PR #924, opened 2025-07-25
Python Wheels: from Tags to Variants [WIP]

mgorny:
Sharing the current draft, so that others can see where this is going. As suggested, I've focused more on the "history" angle of how we arrived at the design rather than precise technical details.

## Text styling

## Non-text contents


apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,401 @@
...
+as flit_core, hatchling, meson-python…) produces a Wheel Variant, which is
+then consumed by a wheel installer (pip, uv, pdm…). Alongside them, variant
+providers (CUDA, ROCm, x86_64, AArch64, BLAS / LAPACK, MPI, OpenMP…) are shown.
+PEP 517 backends communicate with them to validate variants, wheras installers
+do to get supported properties." />
+  <figcaption>Fig. Interaction between wheel software and provider plugins</figcaption>
+</figure>
+</div>
+
+While PEP 517 backends remain responsible for building wheels, and installers
```
rgommers, line 249: Minor editorial comment: I'd use "build backends" as the proper name, and put (PEP 517) in brackets behind it the first time the term build backends is used.  Both because that's the more common name and because there's a push in Python packaging land to refer to things not by a PEP number.

rgommers:
I quite like it so far @mgorny! The narrative style is working well here I think, and the graphics are quite helpful.

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Interesting, I wasn't aware of this. In fact, I've always been using "PEP 517" as a more specific kind of backend. I'll update it.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+<div style={{ textAlign: "center" }}>
+<figure style={{width: 'auto', margin: '0 2em', display: 'inline-block', verticalAlign: 'top'}}>
+  <img src="/posts/python-wheels-from-tags-to-variants/variant-flow-dynamic.png"
+    width="683" height="486" alt="A flow chart showing interaction between
+wheel-related software and dynamic provider plugins. A build backend (such
+as flit_core, hatchling, meson-python…) produces a Wheel Variant, which is
+then consumed by a wheel installer (pip, uv, pdm…). Alongside them, variant
+providers (CUDA, ROCm, x86_64, AArch64, BLAS / LAPACK, MPI, OpenMP…) are shown.
+Build backends communicate with them to validate variants, wheras installers
+do to verify whether the wheels are compatible." />
```
rgommers, line 776: typo "do to verify"

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+So, dynamic plugins can do all that static plugins can do, and more.
+Why do we need two plugin classes then? Why can't we just call it a day,
+say that we've found a better design and go with it? The advantage
+of static plugins is that they are, well, static. Their behavior does not depend
+on the package being installed, they don't need to know what variants
+are available and therefore their results can be reused easily. You just run
+the plugin, snapshot the result and you can reuse it for any package you
+install.
+
+This becomes especially important when we get around to discuss security.
```
rgommers, line 808:
```suggestion
This becomes especially important when we get around to discussing security.
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+
+The discussion around environment markers surfaced more interesting problems.
+For a start, it was pointed out that there could be two different types
+of markers: markers matching the properties of the installed wheel (as outlined
+previously), and markers matching the current system state. For example,
+the latter could be used to install the CUDA plugin if CUDA is available without
+actually having to publish a separate CUDA variant of the main package (think
+JAX). However, this is a recent development and it has not been pursued yet.
+
+It was also pointed out that technically these kind of dependencies can create
```
rgommers, line 942:
```suggestion
It was also pointed out that technically these kind of dependency specifiers can create
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+potential conflicts. For example, since technically a wheel can have multiple
+values for a property, you could end up pulling in two or three versions
+of `nvidia-cuda-runtime` simultaneously. This particularly causes problems
+when creating universal lockfiles — since they need to account for all valid
+combinations of environment markers.
```
rgommers, line 947: It feels like this needs a concluding sentence. Something like _"This is no different from how platform markers work however, and we concluded that it's up to package authors to ensure that there are no overlapping markers"._ Otherwise as a reader I'm taking away the impression that this design doesn't quite work or has an unsolved issue at least.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+
+## Let's talk about security
```
rgommers, line 949: This section feels far too negative, as a reader I'd conclude that this is a hard showstopper. I think this section should say:
1. That we have ideas for how to improve and that the part of the design for this isn't finished yet - it's one of the things that need doing before we can consider submitting a PEP.
2. That currently the _installer_, including all its dependencies (both `pip` and `uv` vendor a whole bunch of them), are currently trusted with higher privileges, and that _variant plugins_ will operate at that same level of higher trust.
3. That that will have implications for how carefully variant plugins are developed, and that "trust on first usage" and "central vetting" are still being considered.

And conclude with something like _"We're confident that this will result in something that will be acceptable to the community after the design and its review are fully complete."_

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+package installation is done with elevated privileges, in order to make
+them available to all users. This makes variant providers a tempting target
+for malicious actors. Three new kinds of supply chain attacks become possible:
+
+1. Injecting malicious code into a new version of an existing variant provider.
+
+2. Injecting malicious code into a new version of the variant provider's
+   (possibly indirect) dependency.
+
+3. Adding a new variant provider that itself is malicious.
```
rgommers, line 980: This one doesn't really do anything unless it starts being used by an existing package.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+
+The issue can be particularly serious, considering that in some setups
+package installation is done with elevated privileges, in order to make
+them available to all users. This makes variant providers a tempting target
+for malicious actors. Three new kinds of supply chain attacks become possible:
+
+1. Injecting malicious code into a new version of an existing variant provider.
+
+2. Injecting malicious code into a new version of the variant provider's
+   (possibly indirect) dependency.
```
rgommers, line 978: Can this be merged into (1)? Just add "or one of its (possibly indirect) dependencies" at the end of (1)?

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+<figure style={{width: 'auto', margin: '0 2em', display: 'inline-block', verticalAlign: 'top'}}>
+  <img src="/posts/python-wheels-from-tags-to-variants/variant-risks.png"
+    width="697" height="373" alt="A diagram showing the three main attack
+vectors at the three additional components of variant wheel's supply chain.
+Firstly, the variant wheel can be attacked directly by adding a malicious provider.
+Secondly, a provider can be attacked by adding malicious code to it. Thirdly,
+one of its dependencies can be attacked in the same manner." />
+  <figcaption>Fig. 10. Variant wheel supply chain attack vectors</figcaption>
+</figure>
+</div>
```
rgommers, line 968: I think I'd like this much better if it showed something like this:
```
Executed at package install time (may have elevated permissions)   |   Executed at time of usage of package/application

Installer + its deps                     | wheel contents
```

going to:

```
Executed at package install time (may have elevated permissions)   |   Executed at time of usage of package/application

Installer + its deps                     | wheel contents--> (arrow to variant plugin)
                                         | 
variant plugin + its deps                |   
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+on a runtime whose version changes frequently, and is not predictable.
+Said version could be `1.0.0`, `1.2.2`, `2.0.10`, `2.3.99`… We're facing
+two problems here. Firstly, the list of all valid values can be very long
+(to be precise, infinitely long). Secondly, we can't fully predict what
+the future versions will be.
+
+Let's start with the lower bound. If the system has version `1.2.2` installed,
+the plugin needs to be declare compatibility with wheels that have a lower bound
+no higher than that: `>=1.2.2`, `>=1.2.1`, `>=1.2.0`… but what is the highest
+value of `>=1.1.n` that we need to support? And what if sometimes upstream
```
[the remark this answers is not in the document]
mgorny: It's decreasing, so 1.2.2, 1.2.1, 1.2.0, 1.1.MAX, ..., 1.1.0 and so on. I'll try to make the problem more explicit.

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Also "wheras".

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+board. It is entirely possible that some of these properties can be solved
+within the existing framework, much like we managed to implement a good enough
+CUDA plugin without having to introduce a dynamic plugin, or that they will
+turn out not to be worth the added complexity. It is also possible that
+some of the existing features will be removed.
+
+There are open issues still. There's the question of how lockfiles are supposed
+to work with variants. There's the possible matter of dependencies on specific
+variants. But most importantly, wheel variants opened a Pandora's box
+of potential security issues that we need to find good mitigations for.
```
rgommers, line 1064: I'm not too enthusiastic about this sentence either:)

mgorny: Yeah, I might have tried to be too metaphoric about it :-).

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+featuredImage:
+  src: /posts/python-wheels-from-tags-to-variants/featured.jpeg
+  alt:
+hero:
+  imageSrc: /posts/python-wheels-from-tags-to-variants/hero.jpeg
+  imageAlt:
+---
+
+# Python Wheels: from Tags to Variants
+
```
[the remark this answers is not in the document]
mgorny: Should we also mention Red Hat here?

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+to facilitate new use cases. To list a few: [PEP 513](https://peps.python.org/pep-0513/)
+added <code>manylinux</code> tags to standardize the dependency on GNU/Linux
+systems, [PEP 656](https://peps.python.org/pep-0656/) added
+<code>musllinux</code> tags to facilitate Linux systems with musl libc.
+
+However, not all new use cases could be handled effectively within
+the framework of tags. The advent of GPU-backed computing made distinguishing
+different acceleration frameworks such as CUDA or ROCm important.
+As many distributions have set baselines for their binary packages
+to x86-64 v2, Python packages also started looking at the opportunity
```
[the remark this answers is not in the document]
mgorny: Familiar with "x86-64-v2" or psABI? If the former, then the Wikipedia page may be easier on the reader: https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+
+Where it gets really interesting is how different projects make use of various
+tag combinations. For example, a single package could combine the following
+conventions:
+
+- `cp311-abi3-manylinux_2_34_x86_64` uses CPython's stable ABI (`abi3`),
+  in the version compatible with CPython 3.11 or newer (`cp311`).
+  It is compatible with GNU/Linux systems with glibc 2.34 or newer,
+  on x86-64 architecture (all that encoded in the `manylinux_2_34_x86_64`
+  platform tag). This is the primary wheel targeting most of the CPython
```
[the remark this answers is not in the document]
mgorny: It meant it's the wheel you're usually want/get. I'll make it say that more plainly.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+  versions.
+
+- `pp311-pypy311_pp73-manylinux_2_34_x86_64` designates a wheel built
+  for PyPy3.11 (`pp311`) versions using `pypy311_pp73` ABI on the same
+  platform. Wheels like these can be used for Python implementations that
+  do not support the stable ABI, such as PyPy or freethreading CPython
+  versions.
```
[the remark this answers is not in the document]
mgorny: Reworded the both ends, please check again on the next push. To be honest, I wanted to showcase the specific idea that you can combine (abi3 + pypy + freethreading) to get good extension coverage, but perhaps it's too detailed.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+let alone that you missed an important step required to get an optimized
+variant. Besides, it is easy to install an incompatible version, two conflicting
+plugins or even reach an unsolvable situation where two different packages you
+need to install have conflicting provider dependencies.
+
+At the same time, plugin discovery was done via [entry
+points](https://packaging.python.org/en/latest/specifications/entry-points/).
+They were quite convenient throughout the development and testing, since they
+enabled us to discover and query installed plugins without having to actually
+interact with any source trees or wheels. However, other found this implicit
```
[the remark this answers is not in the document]
mgorny: I'll say "variantlib" and passive voice.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+</figure>
+
+These markers meant that when the variant wheel was built with a property
+in the listed namespace, the relevant provider plugin would be installed
+in the build environment, and therefore become available to the build backend.
+However, this solution did not last long. It was quite problematic to implement
+properly, since [build](https://pypi.org/project/build/) deferred evaluating
+environment markers to the installer, which implied that we would end up
+with quite a confusing interface — with environment markers in command-line
+arguments processed differently than these coming from packages being installed.
```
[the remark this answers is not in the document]
mgorny: `build` literally passes the contents of `build-system.requires` to pip, i.e.:

```
pip install "cffi>=1.16.0; python_version <= '3.12'" "cffi==1.17.0rc1; python_version >= '3.13'"
```

and so on.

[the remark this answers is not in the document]
mgorny: Perhaps I'm getting too deeply into irrelevant technical details here. I think it's enough to just say it was hard.

apps/labs/posts/python-wheels-from-tags-to-variants.md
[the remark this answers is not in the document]
rgommers: I was thinking that that should go into an Acknowledgements section at the end, that's what I proposed and how we usually do that.

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Because it's not solved yet. The problem is exactly that platform markers are exclusive, whereas variant markers are not. I think "extras" are closer, so maybe that will work as a concluding sentence.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1065 @@
...
+into '*-variants.json' file with the same structure, except that the 'variants'
+objects contains additional 'armv8.2' and 'armv9.0' keys from two other wheels." />
+  <figcaption>Fig. 6. Variant information transit between file formats</figcaption>
+</figure>
+</div>
+
+This is how we arrived at a three step pipeline: with developers putting
+the baseline variant information in `pyproject.toml`, this information being
+then copied to the wheel itself, along with the information about the wheel's
+variant, and eventually the information from all wheel variants being combined
```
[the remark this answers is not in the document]
rgommers: Okay, fine with me, although I personally don't see a reason to worry about that. I'd still rephrase it, since "the wheel's variant" stood out to me - how about " information specifying which variant this wheel is" ?

apps/labs/posts/python-wheels-from-tags-to-variants.md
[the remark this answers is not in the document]
rgommers: Ah okay, I didn't know that, nor got it from that phrasing. I suggest either simplifying the text or spelling it out with an example like this.

apps/labs/posts/python-wheels-from-tags-to-variants.md
rgommers: Maybe conclude with something like "We are still working through some of the finer points of how universal resolving should work" then?

> The problem is exactly that platform markers are exclusive, whereas variant markers are not.

I am not convinced it's all that different; it's easy to cause the exact same problem with platform markers and have an incompatible set of dependency specifiers due to marker overlap. Individual platform markers are exclusive, but combinations of them aren't.

rgommers:
> Okay, I'm going to need to defer bigger changes to the security changes to tomorrow, as my eyes can't take much more today.

Of course. Enjoy the weekend! I'm traveling tomorrow, but might be able to review again on Sunday.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+a single binary package that can be installed everywhere. However, some
+packages are more complex than that — they include compiled Python
+extensions, binaries or even Python code that differs across systems.
+In order to robustly deploy these software on different platforms,
+you need to publish multiple binary packages, with the installers
+being able to select the one best fit the platform used.
+
+For a long time, [Python Wheels](https://packaging.python.org/en/latest/specifications/binary-distribution-format/)
+made do with a relatively simple mechanism
+of providing the needed variance: [Platform Compatibility Tags](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/).
```
zanieb, line 28: "providing the needed variance" is a little awkward / wordy. A minimal change could be "describing the needed variance"?

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+Tags identified different Python implementations and versions,
```
zanieb, line 29: Maybe: "These tags..." to tie it more explicitly to the previous longer name?

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+operating systems, CPU architectures. Over time, they were extended
+to facilitate new use cases. To list a few: [PEP 513](https://peps.python.org/pep-0513/)
+added <code>manylinux</code> tags to standardize the dependency on GNU/Linux
+systems, [PEP 656](https://peps.python.org/pep-0656/) added
+<code>musllinux</code> tags to facilitate Linux systems with musl libc.
```
zanieb, line 34: A little pedantically, I'd say "couple" instead of "few" and use "and" instead of a comma.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+However, not all new use cases could be handled effectively within
+the framework of tags. The advent of GPU-backed computing made distinguishing
+different acceleration frameworks such as CUDA or ROCm important.
+As many distributions have set baselines for their binary packages
+to [x86-64-v2 microarchitecture
+level](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels),
+Python packages also started looking at the opportunity
+to express the same requirement. In fact, the [manylinux_2_34 images are
+blocked on x86-64-v2 becoming the default compiler
+target](https://github.com/pypa/manylinux/issues/1585#issuecomment-3094469339).
```
zanieb, line 45: I was confused coming into this section because I thought it was related to

> The advent of GPU-backed computing made distinguishing different acceleration frameworks such as CUDA or ROCm important.

I think you'll want to do a little more here to make it clear you're enumerating cases.

Part if it might just be the phrasing of this, which doesn't match the style of the previous statement.

A rough example:

> The advent of GPU-backed computing made distinguishing
> different acceleration frameworks such as CUDA or ROCm important.
> Similarly, CPUs have split new features into various microarchitecture 
> levels, such as x86-64-v2. And, numerical libraries support different
> BLAS/LAPACK, MPI, OpenMP providers that are appropriate for different
> users.

You might not want to digress on the `manylinux_2_34` images, or maybe you can put it in a footnote? It distracts from the key message here.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+Numerical libraries support different
+BLAS/LAPACK, MPI, OpenMP providers — and wish to enable the users to choose
+the build using their desired provider.
+While technically tags could be bent to facilitate all these use cases,
```
zanieb, line 49: "While tags could technically be bent..."

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+they would grow quite baroque. Perhaps most importantly, every change
+to tags needs to be implemented in all installers and package-related
+tooling separately, making the adoption difficult.
```
zanieb, line 52: I might combine these sentences, e.g.

> they would grow quite baroque, and, critically, every change ...

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+
+Facing these limitations, software vendors employed different solutions to work
+around the lack of an appropriate mechanism. Eventually, the [WheelNext](https://wheelnext.dev/)
+took up the challenge to design a more robust solution. For the past 6 months we have
```
zanieb, line 56: Maybe: "the WheelNext project"

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+been working hard on “wheel variants”, in callobration
+with [Astral](https://astral.sh/), [NVIDIA](https://www.nvidia.com/)
+and [PyTorch](https://pytorch.org/) release team. This worked resulted
```
zanieb, line 59: "and the PyTorch release team"

Maybe: "This work culminated"

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+in today's [PyTorch 2.8 release](TODO) with experimental variant wheels,
```
zanieb, line 60: It's a little confusing to say "variant wheels" here after "wheel variants" above, while I know you're talking about wheels that use variants, I think this might not be clear to readers. You could just say "experimental new wheels"

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+and a corresponding experimental [variant-capable release of uv package
```
zanieb, line 61: "of the uv ..."

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+</div>
+
+Where it gets really interesting is how different projects make use of various
+tag combinations. For example, a single package could combine the following
+conventions:
+
+- `cp311-abi3-manylinux_2_34_x86_64` uses CPython's stable ABI (`abi3`),
+  in the version compatible with CPython 3.11 or newer (`cp311`).
+  It is compatible with GNU/Linux systems with glibc 2.34 or newer,
+  on x86-64 architecture (all that encoded in the `manylinux_2_34_x86_64`
```
zanieb, line 99:
```suggestion
  on x86-64 architecture (all that is encoded in the `manylinux_2_34_x86_64`
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+
+Well, technically yes. Some use cases would fit well within the current platform
+tags. Say, we could replace the plain `manylinux_2_28_x86_64` tag with a more
+fine-grained `manylinux_2_28_x86_64_v3`. Properties that aren't exclusive
+to existing platforms could be appended to them — say,
+`manylinux_2_28_x86_64_cuda129`, `manylinux_2_28_aarch64_openblas_openmpi`
+and so on. However, there are two major problems with that.
+
+First, the existing [pip](https://pypi.org/project/pip/) code, based
+on the [packaging](https://pypi.org/project/packaging/) library, relies entirely
```
zanieb, line 142: "entirely" feels a bit superfluous here

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+on enumerating all tag combinations compatible with the specific system
+and comparing the tags found in wheels to that list. For example,
+on a GNU/Linux system with Python 3.14 and glibc 2.41, this means almost 1300
+combinations. While this is not that much, once we start adding multiple
+additional suffixes, the number of possible combinations is going to explode.
+Of course, this isn't really a blocker — it is entirely possible to rewrite
+the logic to be more algorithmic and avoid generating all possible combinations.
+However, this is infeasible in practice.
```
zanieb, line 150: It's a bit weird to say it's not really a blocker and then say it's infeasible in practice. This doesn't feel super compelling as written. I might just drop it? If you keep it, I'd maybe say...

> While it's possible to optimize the algorithms that checks for tag compatibility, this becomes is infeasible in practice.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1083 @@
...
+
+Second, tags are implemented entirely within the package manager. This implies
+that for every new set of tags, all package managers must implement them,
```
zanieb, line 153: Maybe "add support for" instead of "implement" so you're not repeating that?

apps/labs/posts/python-wheels-from-tags-to-variants.md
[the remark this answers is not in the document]
mgorny: Actually, I merged your original version, so I guess we're good here :-).

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Okay, done some reshuffling and rewording, and I think I've included all you've suggested. I hope you didn't mean to remove all the paragraphs about how we can mitigate the risks — I think they have the positive tone of "there's a lot of options".

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1095 @@
...
+Language (Python, C++ / Java) and Compute Platform (CUDA 11.8, CUDA 12.6,
+CUDA 12.8, ROCM 6.3, CPU). Below it provides an install command." />
+  <figcaption>Fig. 2. PyTorch version chooser</figcaption>
+</figure>
+</div>
+
+If you enter [PyTorch's “Start Locally”](https://pytorch.org/get-started/locally/)
+document, you are welcomed with an interactive chooser. Once you select
+a specific build, operating system and compute platform, you are given
+a specific pip instruction. Most of these instructions include an `--index-url`
```
zanieb, line 183: Maybe: "pip invocation" or "pip install command"

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Right. I originally thought I'd list more than two, but ended up with two.

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Oh, I've lost the "Initiative" there while rewording.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1092 @@
...
+</div>
+
+There are two main risks being considered here. First, malicious code could
+be introduced into an existing variant provider or one of its dependencies;
+this is a similar risk to how such code could be introduced in the installer
+itself. Second, a new provider could be created that includes malicious code
+by design, and existing packages could start using it.
+
+It can be noted that these risks are similar to Python package build backends,
+that are also downloaded dynamically during the install process and that execute
```
rgommers, line 1017:
```suggestion
It can be noted that these risks are similar to those for build backends,
which are also downloaded dynamically during the install process and execute
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1092 @@
...
+I have already mentioned the possibility of using frozen variant provider output
+instead of running the plugin locally. However, this requires additional effort
+from the user and will primarily  be used in setups with high security
+requirements.
+
+To avoid indirect supply chain attacks, variant providers could recursively pin
+or vendor their dependencies. The latter option is especially interesting, since
+it can be easily enforced at installer level — by forcing provider plugins
+to be installed without dependencies. This way, we mitigate the risk of a new
+version of a compromised dependency being immediately deployed to end users.
```
rgommers, line 1033: Should this also say that we expect most providers to be small and have few or no dependencies? E.g., `nvidia-provider-plugin` depends only on `packaging`, and vendors it in the same way that `pip` does.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1092 @@
...
+pinned to old versions of plugin providers, making it impossible
+to automatically benefit from bug fixes and updates (say, new CPU could fail
+to be recognized as compatible with an old variant).
+
+One interesting option is to maintain a central registry of vetted plugin
+providers. Since such a registry can be updated independently of existing
+wheels, it can mitigate the risk of provider compromise without actually pinning
+to old versions of packages. Unfortunately, the primary problem with such
+a solution is establishing such an authority, and ensuring that it remains
+reliable in the future. It also introduces a single failure point.
```
rgommers, line 1047:
```suggestion
reliable in the future. It also introduces a single point of failure.
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1092 @@
...
+functionality. For example, the `nvidia` plugin needs to be able to access
+the installed NVIDIA libraries and query the GPU.
+
+The more popular provider plugins can also be vendored or reimplemented
+by installers themselves, therefore avoiding reliance on third-party sources.
+However, it assumes that the installer maintainers have to keep track
+of the plugin development, and users cannot benefit from bug fixes
+and improvements without updating the installer.
+
+Finally, there is always the possibility of providing a better control
```
rgommers, line 1061:
```suggestion
Finally, there is always the possibility of providing better control
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1092 @@
...
+
+We're confident that this will result in something that will be acceptable
+to the community after the design and its review are fully complete.
+
+## The present, and the future
+
+In this post, I've attempted to comment on the road the variant wheel work
+has taken from the project's inception to the current pre-proto-PEP: [PEP ### -
+Wheel Variants](http://variant-proposal.wheelnext.dev/).
+Necessarily, it is neither final nor complete in all the detail, but I think
```
rgommers, line 1076:
```suggestion
Necessarily, this story is neither final nor covers all details, but I think
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1092 @@
...
+it does some justice to the complexity of the problem space and the effort
+put into developing a good solution.
+
+Overall, we have been following [the philosophy of WheelNext
+project](https://wheelnext.dev/philosophy_and_design_space/), and focusing
```
rgommers, line 1081:
```suggestion
Overall, we have been following [the philosophy of the WheelNext
initiative](https://wheelnext.dev/philosophy_and_design_space/), and focusing
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1092 @@
...
+on providing a working solutions to real problems. I believe that this approach
```
rgommers, line 1082:
```suggestion
on providing a working solution to real-world problems. I believe that this approach
```

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Added.

zanieb:
Sorry I only made it half way through reviewing it the first time! I'll try to read the second half today.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1133 @@
...
+</figure>
+</div>
+
+The `nvidia` provider detects the local CUDA installation and determines which version
+is installed. Then, it transforms this version into set of compatible
+constraints that can be used to determine whether a given wheel variant
+can be installed — and if multiple variants fit, which one should be used.
+
+For example, if you have CUDA 12.8, then all wheels with lower bound that
+is not higher than that are installable — but a wheel built specifically
```
zanieb, line 317: "all wheels with lower bound that is not higher than that are installable" is pretty confusing. I think you should try to rephrase that in a simpler way.

mgorny: Yeah, I think "that do not require a newer CUDA version" will be clearer.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1144 @@
...
+# Python Wheels: from Tags to Variants
+
+For the past 6 months we have
+been working hard on “wheel variants”, in collaboration
+with [Astral](https://astral.sh/), [NVIDIA](https://www.nvidia.com/)
+and the [PyTorch](https://pytorch.org/) release team. This work culminated
+in today's [PyTorch 2.8 release](TODO) with new wheels utilizing them,
+and a corresponding experimental [variant-capable release of the uv package
+manager](TODO). The user-facing features you can try out today are described
+in [“Wheel variants” by Astral blog](https://astral.sh/blog/wheel-variants),
```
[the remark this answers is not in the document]
mgorny: Hmm, this sounds weird to me. I don't get "this" here, and it's a single blog post, not a blog.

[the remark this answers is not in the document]
rgommers: You had essentially "described in blog", it should be "described in this blog" or "described in a blog". You're missing "a", "the", "this" etc. a fair amount - I think because those don't exist in Slavic languages.

> it's a single blog post

inserting "post" - so "in this blog post" would be better indeed; edited the suggestion to include that.

It'd also be fine to drop the title since it doesn't really add anything, so the link would read something like _this Astral blog post_ or _this blog post by Astral_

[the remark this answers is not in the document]
mgorny: Ah, right. I didn't get that "this" works for "the" there :-). And yeah, I'm trying hard to remember about all the "a"s and "the"s, but this doesn't always work.

As for being more descriptive, I'm trying to follow WCAG, so that all links have descriptive labels that are understandable without context.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+
+<figcaption>Listing 3. Example `pyproject.toml` with sort order defined</figcaption>
+</figure>
+
+However, no reason to stop at namespace. After all, a specific package may
+want to point out, say, that their AES-NI variant is more performant than
+a fallback implemented using the instruction sets provided by x86-64-v3.
+It might also be desirable reorder value; while
+I immediately can't think of a specific use case for that, let's have the option
+for symmetry anyway. So while the plugins specify the initial order of features
```
h-vetinari, line 650: "let's have the option anyway" - not clear to me what is being talked about here. Is this in addition to the toml config just shown, or just describing it after the fact?

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+
+## The null variant
+
+Let's consider that you are distributing a package with three CUDA variants,
+and a CPU variant.
+
+<figure>
+
+```rust
+torch-2.8.0-cp313-cp313-manylinux_2_28_x86_64-cu129.whl
```
h-vetinari, line 665: Previously in the article, the statement was that variants shouldn't be put in the tag, so this is somewhat surprising. Of course, ultimately different wheels need different file names. But what mechanism is supposed to be used for that hasn't been covered I think (or I missed it)

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+torch-2.8.0-cp313-cp313-manylinux_2_28_x86_64-cu126.whl
+torch-2.8.0-cp313-cp313-manylinux_2_28_x86_64-00000000.whl  // CPU-only
+torch-2.8.0-cp313-cp313-manylinux_2_28_x86_64.whl           // CUDA 12.6 + CPU
+```
+
+<figcaption>Listing 5. Example wheel variant filenames, with a null variant and a fallback regular wheel, in order of preference</figcaption>
+</figure>
+
+What we added here is a null variant, with label `00000000`. Since it has
+no properties, it is always supported; that is, as long as variants
```
h-vetinari, line 702: Where is the lack of properties for that variant defined? Surely it doesn't come from a magic string `00000000` (and in the unlikely case that this is so, what about `0000000` or `000000000`?)

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+plugin-api = "provider_variant_aarch64.plugin:AArch64Plugin"
+
+[variant.providers.x86_64]
+requires = ["provider-variant-x86-64 >=0.0.1,<1"]
+# use only on x86_64 machines
+enable-if = "platform_machine == 'x86_64' or platform_machine == 'AMD64'"
+plugin-api = "provider_variant_x86_64.plugin:X8664Plugin"`
+```
+
+<figcaption>Listing 3. Example `pyproject.toml` with sort order defined</figcaption>
```
h-vetinari, line 642: This doesn't cover how package authors would assign priorities between variants, e.g. whether to prefer higher CUDA or higher CPU optimizations first. The text makes it appear like that should be possible, but it's quite unclear how that would look like (and also how a user would be able to override any of this, TBH).

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+
+```toml
+dependencies = [
+    'triton==3.4.0; platform_system == "Linux" and "nvidia" in variant_namespaces',
+    'nvidia-cudnn==9.10.2.21; platform_system == "Linux" and platform_machine == "x86_64" and "nvidia" in variant_namespaces',
+    'nvidia-cusparselt==0.7.1; platform_system == "Linux" and platform_machine == "x86_64" and "nvidia" in variant_namespaces',
+    'nvidia-nccl==2.27.3; platform_system == "Linux" and platform_machine == "x86_64" and "nvidia" in variant_namespaces',
+
+    # CUDA 12.6
+    'nvidia-cuda-nvrtc==12.6.77; platform_system == "Linux" and platform_machine == "x86_64" and "nvidia :: cuda_version_lower_bound :: 12.0" in variant_properties',
```
h-vetinari, line 963: Is the divergence between 12.6 and 12.0 here intentional?

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+For a start, it was pointed out that there could be two different types
+of markers: markers matching the properties of the installed wheel (as outlined
+previously), and markers matching the current system state. For example,
+the latter could be used to install the `nvidia` plugin if CUDA is available without
+actually having to publish a separate CUDA variant of the main package (think
+JAX). However, this is a recent development and it has not been pursued yet.
+
+It was also pointed out that these kind of dependency specifiers can create
+potential conflicts. For example, since technically a wheel can have multiple
+values for a property, the code in listing 8. could end up pulling in two or three versions
```
h-vetinari, line 999: Stale reference. I think this should be listing 9

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+is that for a variant to be compatible, all its features had to have a value
+compatible with the system in question. This logic is conjunctive, that is representing
+logical `AND`. You need to have a supported CPU, and a CUDA version that is
+not older than the lower bound, and a CUDA version (the same one) that is older
+than the upper bound.
+
+CPU instruction sets are an interesting case. Each
+instruction set is represented by a separate feature, with a single possible
+value of `on`. If the user's CPU supports a given instruction set, it supports
+said `on` value; if it doesn't, the feature has no compatible values. A wheel requiring
```
h-vetinari, line 729: This is perhaps too niche to go further into detail, but I'm left wondering if this will now by default create enormous list of cpu features in variant_features, or whether this is talking about a separate provider than the x64 v1-v4 aggregates

h-vetinari (commented):
Gave this a read through as an outsider to the initiative. Perhaps the feedback is useful. :)

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+
+The `nvidia` provider detects the local CUDA installation and determines which version
+is installed. Then, it transforms this version into set of compatible
+constraints that can be used to determine whether a given wheel variant
+can be installed, and if multiple variants fit, which one should be used.
+
+For example, if you have CUDA 12.8, then all wheels that do not require a newer
+CUDA version are installable, but a wheel built specifically
+for CUDA 12.8 will be preferred. So we support wheels with a lower bound
+of `12.8`, `12.7`, `12.6`, `12.5`… The upper bound is handled in a similar way.
```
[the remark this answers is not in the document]
mgorny: I'm confused. But the paragraph before it is talking about the provider, and this is the continuation of it?

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Yep, 12.6 builds are used for `>=12.0,<12.8`. Perhaps I should make it clear above.

apps/labs/posts/python-wheels-from-tags-to-variants.md
```diff
@@ -0,0 +1,1146 @@
...
+This structure is transferred to a '*.dist-info/variant.json' file in the wheel,
+with the same objects as subtables of the 'variant' table, plus an additional
+'variants' object with a single 'x8664v3' subobject. This is in turn converted
+into '*-variants.json' file with the same structure, except that the 'variants'
+objects contains additional 'armv8.2' and 'armv9.0' keys from two other wheels." />
+  <figcaption>Fig. 6. Variant information transit between file formats</figcaption>
+</figure>
+</div>
+
+We arrived at a three step pipeline: developers put
```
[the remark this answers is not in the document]
mgorny: To be honest, this post wasn't really meant as a description of the "current state", but rather a curious read of how things changed over time. But I'll try to clarify this a bit. That said, I can't think of much I could do here.

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Uh, I feel like I've been switching between past to present tense too much.

apps/labs/posts/python-wheels-from-tags-to-variants.md
mgorny: Yeah, I think that's going too deeply into implementation details. Here the focus is on the possibility of defining properties like that. And well, technically speaking you could do either (or both), but our recommendation would be to do baseline + stuff above baseline.

h-vetinari:
> Could you give it another look?

Looks like a nice improvement to me!

mgorny:
Thanks. I've pushed some cosmetic changes (using "listing N." instead of "following" for consistency), and I think it's ready to be merged at the appointed time.

PR #958, opened 2026-03-12
Upgrade dependencies to latest versions

lysnikolaou:
Upgrade to Nx 22 with React 19 and TypeScript 5.7

Nx migrate from 13.8.5 to 22.5.4, which also bumps:
- React 17.0.2 → 19.2.4 (with @types/react@19)
- TypeScript 4.5.2 →5.7.3
- Next.js 12.2.5 → 16.1.6
- Jest 27 → 30 (+ jest-environment-jsdom)
- @testing-library/react 12 → 16

To facilitate this, the following configuration changes:
- jest.config.js becomes jest.config.ts
- Package names change from nrwl to nx
- nx.json automatic migration with the nx migrator
- project.json's were slightly changed - no important changes

Code compatibility:
- Add PropsWithChildren to LinkWithArrow component
- Fix storyblok-react SbEditable types for React 19 children
- Wrap Page component array return in Fragment
- Fix Link companent with anchor children - no longer needed

Stylistic changes:
- As far as I can see there's only one visible stylistic change, the blog post categories render a bit bigger than they used to. Open for discussion whether this is okay or not. If not, I can change the code to render as they were before.

Generated-By: Claude Opus 4.6 <noreply@anthropic.com>

 ---

This has been my weekend side project for a few weeks now. I feel it brings the project to a state where things are up-to-date and new code changes and/or bigger projects can be worked on without having to fight old dependencies behaving weirdly. Bonus points are that:
- Code vulnerabilities went down from 108 to 23. None critical anymore.
- Build time appears to be better. 
- Bundle size appears to be slightly optimized but I haven't done a lot of testing on that.

I did this with the help of automatic migrator tools, as well as Claude Code. I used planning mode and instructed Claude Code to use the tools to do a migration step-by-step (Nx 13->14->15, etc.). Every time it would change some configuration as part of the migration, I'd ask what exactly the configuration value does, why it was changed and to give me a reference to the CHANGELOG that explains the change. Most of the time, it gave me the correct CHANGELOG, sometimes I had to find it myself. I also instructed it to run the server (for manual testing) and build a production bundle after each step to verify things continue working. Even with all of this in place, I manually verified the every line of code changed/added/removed is correct, so I'm pretty confident that this is good to go. 

I also have three follow-up PRs that migrate to new versions of ESLint, @storyblok/react and next/image. I'll open those after (if) this has been merged.


lysnikolaou:
Thanks for the having a look at this @rgommers! Worked through each item:

Must fix
1. Post/RSS paths - verified correct. Nx 22's @nx/next executors run with cwd at project root, so the apps/labs/ prefix had to go.
2. typescript-eslint v8 - removed; the v7 packages are what .eslintrc.json actually uses.
3. Lib lintFilePatterns - verified unneeded; the schema default `[{projectRoot}]` discovers files.

Should fix

5. Removed `experimental.images.allowFutureImage` from consulting.
6. Root jest.config.ts now uses `import`.
7. Empty `globals: {}` removed from the 3 lib jest configs.

Nice to have

8. Deferring next/legacy/image migration to a follow-up PR.
9. `x-forwarded-for` is correct for Edge; I've verified it works correctly.
10. `bordel-l-white` typo fixed.

lysnikolaou:
> I did some browsing of the preview and the only issue I noticed is that the blog post categories are rendered in a much larger font, which looks off.

Fixed this as well. Turns out some Tailwind change affected this.

rgommers (approved):
This all seems to work as advertised in the preview, and looks like all the important items were taken care of. With CI green, let's get this in - thanks again @lysnikolaou for finishing this up!

PR #969, opened 2026-05-06
Migrate from next/legacy/image to next/image

lysnikolaou:
Replaces all `next/legacy/image` imports with the modern `next/image`
across both apps and the shared UI library. The Picture wrapper keeps
its existing API (layout/objectFit/objectPosition) and translates legacy
props internally, so all 30+ call sites remain unchanged. `Hero`,
`HeroResponsiveImages`, and `LinkWithArrow` are converted directly.

In the default (intrinsic) branch of Picture, an explicit CSS
`aspect-ratio` is set to preserve the legacy behavior where e.g. team images
always render at exactly 200x200 regardless of source dimensions.

Generated-by: Claude Opus 4.7

### DIFF
--- a/apps/consulting/types/storyblok/bloks/hero.ts
+++ b/apps/consulting/types/storyblok/bloks/hero.ts
@@ -1,9 +1,8 @@
-import { ImageProps } from 'next/legacy/image';
-
 import { TImage, TBlok } from '@quansight/shared/types';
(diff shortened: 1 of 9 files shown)


rgommers (approved):
LGTM, thanks @lysnikolaou!

PR #972, opened 2026-05-26
[BLOG] add 'Scaling NumPy on Free-Threaded Python` post

kumaraditya303:
## Text styling

## Non-text contents


apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+on the same machine.
+
+The free-threaded build removes the GIL, but removing the GIL is not enough on
+its own. Profiling the reproducer revealed several hidden bottlenecks in NumPy
+and CPython that were masked by the GIL on the regular build but became severe
+scaling issues on the free-threaded build. These scaling bottlenecks fall
+primarily into three categories:
+
+1. Lock contention.
+2. Refcount contention on shared global objects.
```
ngoldbaum:
```suggestion
2. Reference count contention on shared global objects.
```

A little less jargony, although you explain it later.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+    style={{position:'relative'}}
+  />
+  <figcaption>Flame graph of the reproducer before any fixes.</figcaption>
+</figure>
+
+## The bottlenecks and the fixes
+
+### 1. Lock contention in `tracemalloc`
+
+The first bottleneck that surfaced was in `tracemalloc` of CPython.
```
ngoldbaum:
```suggestion
The first bottleneck that surfaced was in CPython's `tracemalloc` module.
```

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+`tracemalloc` is a memory tracking tool which is disabled by default,
+however even though it was not enabled, it still acquired a global lock
```
ngoldbaum:
```suggestion
`tracemalloc` is a memory tracking tool which is disabled by default.
However even though it was not enabled, it still acquired a global lock
```

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+on every allocation and deallocation.
```
ngoldbaum:
```suggestion
on every allocation and deallocation to check whether `tracemalloc` had been
enabled at runtime.
```

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+
+I fixed this in CPython by avoiding locking when `tracemalloc` is disabled
```
ngoldbaum:
```suggestion
I fixed this in CPython by avoiding locking when `tracemalloc` is disabled.
```

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+by using atomic operations to check whether it is enabled before acquiring
+the lock. This was implemented in
+[python/cpython#143065](https://github.com/python/cpython/pull/143065).
```
ngoldbaum:
```suggestion
Now, CPython uses atomic operations to check whether `tracemalloc` is
enabled, providing a lock-free fast path. This was implemented in
[python/cpython#143065](https://github.com/python/cpython/pull/143065).
```

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+
```
ngoldbaum, line 94: Maybe worth adding a paragraph here explaining how this sort of bug is easy to write and that it's often not obvious at first whether or not a lock will cause a scaling issue. Also that adding a fast path is usually how you fix scaling issues like this.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+### 2. Lock contention in the ufunc dispatch cache
+
+Every call to a ufunc such as `np.sin` or `np.cos` goes through a dispatch
+cache that maps a tuple of input types to a concrete loop implementation. The
+cache was previously protected by a `std::shared_mutex` (a reader-writer lock).
+Even though, it was a reader-writer lock, it still didn't scale well.
```
ngoldbaum: I would break this into two paragraphs. The first could be something like this:

> NumPy implements mathematical operations like `np.sin` as ufuncs (short for universal functions). You can read more about ufuncs in the NumPy docs [add link]. The key thing to know for this discussion is that that  ufuncs are defined in terms of the input types. NumPy implements different loops for different types. For example, the implementation for `np.add` is different when the operands for the addition operation are integers or floats. For the former, NumPy needs to use a loop based on *integer* addition while for the latter we need to use floating point multiplication. In other words, NumPy has a dispatch system where it needs to determine the correct loop given a tuple of input data types.

And then the second could explain that there's a cache that speeds this up and that the cache is global state in the NumPy implementation.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+
+I fixed this by implementing a lock-free hashtable for the dispatch cache,
+which allows for lock-free reads. The fast-path for cache hit is now just
+a few atomic read operations which scales well across threads. This was
+implemented in
+[numpy/numpy#30593](https://github.com/numpy/numpy/pull/30593), you can
+read more about the lock-free design in
+[npy_hashtable.c](https://github.com/numpy/numpy/blob/main/numpy/_core/src/common/npy_hashtable.c).
+
+TODO: describe lock-free implementation in more detail and add a diagram.
```
ngoldbaum: Awesome, looking forward to reading this.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+
+### 3. Reference count contention on global `PyCapsule` objects
+
+NumPy stores the default memory handler in a global `PyCapsule` object.
```
ngoldbaum, line 176: Similar to above, add a paragraph above this explaining why NumPy has a memory handler than lives in a PyCapsule: to allow customization.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+On the GIL build, an extra `Py_INCREF`/`Py_DECREF` on these capsule objects
+is essentially free. On the free-threaded build, every increment and
+decrement of such objects requires atomic operations on the reference count,
+and with many threads all contending on the same global capsules, this
+became a severe scaling bottleneck.
+
```
ngoldbaum: Maybe add a paragraph here or link elsewhere to explain why incref and decref is *so* much slower for shared objects on the free-threaded build and how global state is particularly susceptible to triggering slowdowns due to reference count churn.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+I fixed this by making the global default memory handler immortal,
+meaning that it is never freed and no reference counting operations are
+performed on them at all. There was no public C API to mark an object as
+immortal, so I added `PyUnstable_SetImmortal` in CPython. This was
+implemented in
+[python/cpython#144543](https://github.com/python/cpython/pull/144543)
+and [numpy/numpy#30826](https://github.com/numpy/numpy/pull/30826)
```
ngoldbaum: Consider breaking this into two paragraphs where the first describes what making an object immortal means and that CPython uses this a lot internally to allow things to scale better. End by explaining how before Python 3.15, there was no public API for libraries like NumPy to make things immortal. End the second paragraph by explaining how this use-case justified adding the API and that we can make this immortal on Python 3.14 by using the compat headers, which implement the operation using private C APIs with the knowledge that Python 3.15 made the functionality public.

Also consider adding a paragraph about free-threading is still new and there still might be things that are needed in the C API to allow extensions to be performant. This example shows how it's not a lost cause to get the performance win, even on released Python versions, so long as you go through the trouble to expose the functionality you want in the CPython development branch.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+
+### 4. Module attribute lookup contention
+
+NumPy uses module level `__getattr__` to resolve ufuncs such as `np.sin`
+and `np.cos` to their actual implementations. In CPython, the module
```
ngoldbaum: Consider adding a sentence spelling out a little more concretely that you mean literally the attribute access on the `numpy` module.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+attribute lookup bytecode specialization was not enabled for modules
+which defined `__getattr__` which causes the attribute lookup to follow
+the slow-path of acquiring the import lock and performing the lookup.
+
+I fixed this upstream in CPython
+([python/cpython#143470](https://github.com/python/cpython/pull/143470))
+by enabling bytecode specialization for module attribute lookups even
+when `__getattr__` is defined.
```
ngoldbaum: Maybe say that this fix is only available in Python 3.15 (right?) but if you hit this in 3.14, you can avoid the slowdown by avoiding doing `getattr` repeatedly in a hot loop.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+
+<figure style={{ textAlign: 'center' }}>
+  <img
+    src="/posts/scaling-numpy-on-free-threaded-python/before.png"
+    alt="Flame graph of the reproducer before any fixes."
+    style={{position:'relative'}}
+  />
+  <figcaption>Flame graph of the reproducer before any fixes.</figcaption>
+</figure>
+
```
ngoldbaum, line 79: Consider adding a section here explaining how to read flame graphs (or linking somewhere that does) and explaining how you actually debug stuff like this. Something like "first, run the script under samply, look at a compute thread, and try to find things in the flame graph that are taking a substantial fraction of runtime compared with calling a low-level `sin` or `cos` implementation, then try to fix the issue, and run samply again".

Maybe also talk about how you can compare with the multiprocessing version in the same samply output and this also helps spot things that are causing overhead.

Also worth emphasizing that after you fix a scaling bottleneck, that will only improve scaling until the next-worse bottleneck is fixed, so this is a very iterative process and it's not necessarily unusual that there were several different issues that all needed to be fixed to get close-to-ideal scaling.

Also maybe worth talking a little about Amdahl's Law and how any time spent running on only one core (e.g. with a global lock held) necessarily means that there is an ultimate limit to multithreaded scaling.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+After all of the above, the only remaining contention in the flame graph was
+inside `malloc` itself. The free-threaded workload was allocating memory
+concurrently across many threads, and it didn't scale well because of
+locking inside the system allocator implementation particularly on macOS.
+
+The fix was twofold:
+
+- In CPython, I changed the raw allocator APIs to use mimalloc as the
+  underlying memory allocator on the free-threaded build. This was implemented
+  in [python/cpython#144916](https://github.com/python/cpython/pull/144916).
```
ngoldbaum, line 241: Add a sentence explaining how `mimalloc` is designed to have good multithreaded scaling and this completely bypasses poor multithreaded in scaling on system allocators like the one on MacOS.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+
+- In NumPy, I changed the array allocation APIs to use the raw allocator
+  instead system allocator. This was implemented in
```
ngoldbaum:
```suggestion
  instead of the system allocator. This was implemented in
```

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+concurrently across many threads, and it didn't scale well because of
+locking inside the system allocator implementation particularly on macOS.
+
+The fix was twofold:
+
+- In CPython, I changed the raw allocator APIs to use mimalloc as the
+  underlying memory allocator on the free-threaded build. This was implemented
+  in [python/cpython#144916](https://github.com/python/cpython/pull/144916).
+
+- In NumPy, I changed the array allocation APIs to use the raw allocator
```
ngoldbaum: I would maybe add a whole paragraph explaining how NumPy called `malloc` and `free` directly but that CPython offers low-level "raw" allocator routines NumPy could use instead. Doing the change to using the raw allocator allows NumPy to use `mimalloc` instead of the system allocator without knowing anything about `mimalloc`. Also add a link to the CPython memory allocation docs.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+    style={{position:'relative'}}
+  />
+  <figcaption>
+    The red line represents the performance before the fixes, the green line
+    represents the performance after the fixes and the black line represents
+    the performance of the multi-process version.
+  </figcaption>
+</figure>
+
+Before the fixes, the multi-threaded scaled well upto 18 threads, but after
```
ngoldbaum:
```suggestion
Before the fixes, the multi-threaded case scaled well up to 18 threads, but after
```

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,180 @@
...
+that because of the bottlenecks described above, the performance degraded
+significantly and became much slower than the multi-process version. After
+the fixes, the multi-threaded version scales well across all 32 cores and is
+significantly faster than the multi-process version.
+
+## Summary
+
+NumPy now scales well on the free-threaded build of CPython after several
```
ngoldbaum:
```suggestion
NumPy ufuncs now scale well on the free-threaded build of CPython after several
```

ngoldbaum (commented):
Looks great! Lots of mostly nitpicky suggestions inline as well as a few more substantive suggestions.

rgommers:
The hero image might be better with a light background? This looks a little off:

image

rgommers:
Had a quick read, this post has a really nice story line. And I quite like the diagram for the lock-free ufunc dispatch cache.

kumaraditya303:
> The hero image might be better with a light background? This looks a little off:

I changed it to a lighter colour now, thanks for the suggestion!

rgommers:
Is this still Draft, or ready for review?

kumaraditya303:
> Is this still Draft, or ready for review?

This is ready for review, I kept it in draft because I don't want it to be published yet.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
ngoldbaum: On second thought this content can live elsewhere.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,269 @@
...
+    style={{position:'relative'}}
+  />
+  <figcaption>
+    The lock-free dispatch cache. Readers follow the atomic pointer to the
+    current buckets table and look up entries without taking a lock. After the
+    table is resized, the old table is kept alive (linked through a prev
+    chain) until deallocation, so readers still using it — like Reader C — can
+    finish safely. Only writers acquire the mutex.
+  </figcaption>
+</figure>
```
ngoldbaum: I think the explanation above makes more sense if you move this diagram up one or two paragraphs.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,269 @@
...
+The problem was that, before Python 3.15, there was no public C API for a
+library like NumPy to make its own objects immortal. This use case is exactly
+what justified adding one, so I added `PyUnstable_SetImmortal` in CPython. This
+was implemented in
+[python/cpython#144543](https://github.com/python/cpython/pull/144543)
+and [numpy/numpy#30826](https://github.com/numpy/numpy/pull/30826). The API is
+public from Python 3.15 onwards, but NumPy can already use it on Python 3.14
+through the [pythoncapi-compat](https://github.com/python/pythoncapi-compat)
+headers, which implement it using the private C APIs that back the public
+function in 3.15.
```
ngoldbaum: 3.14, right?

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,269 @@
...
+
+### 4. Module attribute lookup contention
+
+When you write `np.sin`, that is literally an attribute access on the `numpy`
```
ngoldbaum: I'd drop "literally"

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,269 @@
...
+system allocator does not scale well when multiple threads are allocating and
+deallocating memory.
+
+CPython exposes its own low-level ["raw" memory allocator
+routines](https://docs.python.org/3/c-api/memory.html) such as
+`PyMem_RawMalloc` and `PyMem_RawFree`. However, those routines were also
+calling the system allocator. The free-threaded build of CPython already uses
+[mimalloc](https://github.com/microsoft/mimalloc), which is highly optimized
+for multi-threaded workloads, to allocate Python objects, but NumPy was not
+using it because it was calling the system allocator directly.
```
ngoldbaum: Break this run-on up into more sentences.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
```diff
@@ -0,0 +1,269 @@
...
+  underlying memory allocator on the free-threaded build. This was implemented
+  in [python/cpython#144916](https://github.com/python/cpython/pull/144916).
+
+- In NumPy, I changed the array allocation APIs to use CPython's raw allocator
+  instead of calling the system allocator directly. Routing array allocations
+  through these routines lets NumPy automatically benefit from mimalloc —
+  without needing to know anything about mimalloc itself. This was implemented
+  in [numpy/numpy#30846](https://github.com/numpy/numpy/pull/30846).
+
+## Benchmarks
```
ngoldbaum, line 249: This is the same benchmark from stackoverflow or something else? Consider adding a sentence or two explaining the workflow you're benchmarking.

ngoldbaum (approved):
LGTM! Really nice story. This could certainly be a conference talk if you find one you'd like to go to in the next year or so. Just a thought.

I spotted a few minor issues that I left comments on inline.

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
kumaraditya303: done

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
kumaraditya303: yes

apps/labs/posts/scaling-numpy-on-free-threaded-python.md
kumaraditya303: yes, added

kumaraditya303:
> I spotted a few minor issues that I left comments on inline.

I've addressed them and did a minor editing pass for the final fixes. 

@rgommers I'd like this to be published on 29th of this month.

rgommers:
Sure that works for me; I set myself a reminder for the 29th. This looks like it's ready to go

rgommers:
Okay, let's get this up, thanks Kumar! And thanks Nathan for the review!

PR #979, opened 2026-06-08
Add Python ABI blog post

ngoldbaum:
This blog post contains content that will go into [my EuroPython talk](https://ep2026.europython.eu/session/what-every-python-developer-should-know-about-the-cpython-abi).

I'm adding it as a blog post to help me structure my thoughts for the talk and so marketing has a post to share about this topic.

## Text styling

## Non-text contents


apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,175 @@
...
+To access the internals of the interpreter, one needed only to `#include "Python.h"` in a C program, which gave rich access to the internal state of the interpreter and hooks to call into the interpreter via the CPython C API, the very same C API that the core interpreter implementation is based on.
+
+What is the C API exactly?
+It's precisely the set of C macros, typedefs, functions, and structs exposed by the header that defines the C interface: `Python.h`
+This constitutes an enormous number of symbols that collectively allow developers to write code in C that allows similar functionality to any arbitrary Python script.
+The [Cython](https://cython.readthedocs.io/en/latest/) programming language takes advantage of this by (among other things) literally compiling Python code to a C "script" that when linked into a larger program behaves identically to the Python code it was compiled from.
+Python JITs like [Numba](https://numba.pydata.org), [torch.compile](https://docs.pytorch.org/tutorials/intermediate/torch_compile_tutorial.html), and other third-party Python JITs do a similar trick at runtime, also via the C API.
```
rgommers: This sentence seems misleading. Numba doesn't generate C, it does bytecode analysis and then goes to LLVM IR. `torch.compile` hooks in at the frame evaluation level, also pokes at bytecode, and then targets (depending on backend) something PyTorch-specific, like the PyTorch C++.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,175 @@
+---
+title: 'What Every Python Developer Should Know About the CPython ABI'
+authors: [nathan-goldbaum]
+published: June 4, 2026
+description: 'An introduction to the concept of the Application Binary Interface (ABI), the various CPython ABIs, and the new abi3t stable ABI in Python 3.15.'
+category: [PyData ecosystem]
+featuredImage:
+  src: /posts/python-abi-abi3t/puzzle.png
```
rgommers: This intuitively felt wrong to me as soon as I glanced at it. In my mind the chain is `.py` --- import machinery ---> finds extension module ---> extension module does its thing via the CPython ABI.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,175 @@
...
+Notably `abi3` wheels are not installable on [the free-threaded build](https://py-free-threading.github.io) of CPython.
+We'll see below why this is the case and how the new `abi3t` ABI in Python 3.15 will help ameliorate this limitation going forward.
+
+### What is the Python ABI?
+
+Referring to only one Python ABI is a bit of a fib.
+Each [target triple](https://mcyoung.xyz/2025/04/14/target-triples/) - the combination of CPU architecture, vendor, and operating system - has its own distinct ABI. However, one can logically group all the platform-specific details of a specific ABI into its own bucket and all the Python-specific details into another bucket, which I informally refer to as "the" Python ABI.
+In practice, the Python ABI includes all of the variables and function declarations exposed in `Python.h`.
+Here, a function definition is the name of the function, the number of arguments, the types of the arguments, and the type of the value returned by the function, if any.
```
rgommers: You're using the word "symbol" in two places, but it's missing a definition and it's connected to what's in this sentence (this is the "implementation" behind the symbol name I'd say). That seems useful to clarify somehow, including that this is all implicit except for function name <-> symbol name.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,175 @@
...
+
+You may wonder: why go into all this detail on this struct? Because the layout of `PyObject` is one of the main sources of tension that led to the development of two new python ABIs in recent years. We'll get more into the history and future of the Python ABI below.
+
+## The Past and Future of the Python ABI
+
+### The Version-Specific ABI and the Python 2 -> 3 transition
+
+In the early days, there wasn't much distinction between "the interpreter" and "what's exposed by the C API": one simply was able to monkey around at will with internal state in the interpreter.
+While this enabled lots of cool stuff, it also had a cost: extensions regularly broke with different Python releases, requiring careful fixes to adapt to changes in interpreter internals.
+This also introduced a design pressure to freeze internals, lest people building on Python complain about changes breaking their code.
```
rgommers, line 187: Will this section, or some other section, explain public vs. private vs. internal? I think that 3-way split is kind of unusual and the audience would benefit from a clear explanation.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,175 @@
...
+Taken together, these two pieces of information are always tracked on every Python object. In some sense, the object _is_ the reference count, type, and the address of the `PyObject` struct.
+
+To make that all a little more concrete, `object()` in Python instantiates a `PyObject` instance in ther interpreter runtime, while `dict()` instantiates a `PyDictObject` struct - a different struct that extends `PyObject`.
+That is, the first two fields of `PyDictObject` are exactly the same as `PyObject`: all Python objects correspond either to an instance of `PyObject` or a struct that extends `PyObject.
+
+You may wonder: why go into all this detail on this struct? Because the layout of `PyObject` is one of the main sources of tension that led to the development of two new python ABIs in recent years. We'll get more into the history and future of the Python ABI below.
+
+## The Past and Future of the Python ABI
+
+### The Version-Specific ABI and the Python 2 -> 3 transition
```
rgommers: Is 2 -> 3 transition relevant? A lot of stuff broke, but I'm not sure it matters that much, and it's treading into territory people have lots of strong opinions on.

apps/labs/posts/python-abi-abi3t.md
rgommers: Or actually, you could explain the whole thing more carefully as a five-way split perhaps?

1. Limited
2. Public
3. Unstable
4. Private
5. Internal

rgommers (commented):
Looks like a good start - I had some time, so did a read through. My comments focus only on the main conceptual things that stood out to me.

ngoldbaum:
I did another pass today to address Ralf's comments and try to fill in the rest of the text.

It's not quite done yet but hopefully I can finish this off next week. If anyone would like to read this over and give me comments, please feel free.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+hero:
+  imageSrc: /posts/python-abi-abi3t/cpython_api_layers_hero.png
+  imageAlt: 'Five nested ellispes illustrating the layering of the Python C API. The outermost ellipse is gray and labeled "Internal API". The next enclosed ellipse is red and is labeled "Private API`. The next enclosed ellipse is yellow and is labeled "Unstable API". The next enclosed ellipse is blue and labeled "Version-specific API". The next enclosed ellipse is green and is labeled "Limited API".'
+---
+
+# What Every Python Developer Should Know About the CPython ABI
+
+The CPython Application Binary Interface (ABI) backs Python's main superpower: the ability to easily call into native C, C++, Rust, or Fortran code and for that code to call back into the interpreter and update the state of Python objects.
+But what exactly is an ABI? How does it differ from an Application Programming Interface (API)?
+Why did I begin this post's the title with "What _Every_ Python Developer Should Know"?
```
JelleZijlstra:
```suggestion
Why did I begin this post's title with "What _Every_ Python Developer Should Know"?
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+Via the [CPython C API](https://docs.python.org/3/c-api/index.html) and the corresponding [Python ABI](https://docs.python.org/3/c-api/stable.html#c-api-stability).
+But what does that mean exactly and what even is a C API and how does it differ from the ABI?
+It certainly doesn't help things that the two related and intertwined concepts are referred to using similar-sounding acronyms.
+
+[CPython](https://github.com/python/cpython), as the name suggests, is implemented in the C programming language.
+It began as a research project and was written to be easy to extend with new functionality.
+To access the internals of the interpreter, one needed only to `#include "Python.h"` in a C program, which gave rich access to the internal state of the interpreter and hooks to call into the interpreter via the CPython C API, the very same C API that the core interpreter implementation is based on.
+
+What is the C API exactly?
+It's precisely the set of C macros, typedefs, functions, and structs exposed by the header that defines the C interface: `Python.h`
```
JelleZijlstra:
```suggestion
It's precisely the set of C macros, typedefs, functions, and structs exposed by the header that defines the C interface: `Python.h`.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+This constitutes an enormous number of symbols.
+It's possible to write code in C that allows similar functionality to any arbitrary Python script at the cost of compilation, verbosity, and exposure to the pitfalls of the C programming language.
+The reward is raw executation speed.
```
JelleZijlstra:
```suggestion
The reward is raw execution speed.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+It is often possible to achieve order-of-magnitude or even several order-or-magnitude size speedups by translating Python code to a compiled language that can call into the C API.
+The [Cython](https://cython.readthedocs.io/en/latest/) programming language takes advantage of this by (among other things) literally compiling Python code to a C "script" that when linked into a larger program behaves identically to the Python code it was compiled from.
```
JelleZijlstra: Not always completely identically of course; not worth going into in too much detail but the current wording feels too strong.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+What isn't the C API?
+The C API is purely a construct of the C programming language.
+Code written in languages that aren't C can call into a C API, but only be using the conventions of the C programming language and abstracting away functionality that is not expressible in C.
```
JelleZijlstra:
```suggestion
Code written in languages that aren't C can call into a C API, but only by using the conventions of the C programming language and abstracting away functionality that is not expressible in C.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+A platform ABI determines things like the exact way machine code needs to pass arguments to a function. For example, on the [`x86_64`](https://wiki.osdev.org/System_V_ABI#x86-64) architecture, only the first six non-floating-point arguments of a function are passed via registers, the remaining arguments are passed via the stack.
+Normally, developers do not need to worry about details like this: compilers automatically generate machine code appropriate for whatever platform ABI a developer wants to target.
+However, it _is_ important to know that different operating systems and CPU architectures have unique ABIs and that code compiled for one platform ABI is completely unusable on another platform ABI.
+This fact determines much of the design of the [binary wheel distribution format](https://packaging.python.org/en/latest/specifications/binary-distribution-format/), which we'll discuss in more detail below.
+
+The biggest consequence is that each platform and CPU architecture, each with its own distinct platform ABI, requires their own unique builds.
```
JelleZijlstra:
```suggestion
The biggest consequence is that each platform and CPU architecture, each with its own distinct platform ABI, requires its own unique builds.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+This is one reason projects like `NumPy` distributes so many binary wheels with each release: projects need to build binaries for each platform ABI they want to support.
```
JelleZijlstra:
```suggestion
This is one reason projects like `NumPy` distribute so many binary wheels with each release: projects need to build binaries for each platform ABI they want to support.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+ <figure style={{ textAlign: 'center' }}>
+   <img
+     src="/posts/python-abi-abi3t/wheel-filename-anatomy.png"
+     alt="A diagram illustrating the structure of a wheel filename. There are six pieces of information as part of each wheel file: the name of the distribution, the distribution's version, the Python version tag, the abi tag, and the platform tag, and the wheel filename extension."
+     style={{position:'relative',left:'0%',width:'100%'}}
+   />
+ </figure>
+
+The pieces that are interesting for ABI compatibility are the third, fourth, and fifth tag.
+These are the [compatbility tags](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/) for the wheels. The "python" tag indicates either the exact version supported by the wheel or the minimum supported version.
```
JelleZijlstra:
```suggestion
These are the [compatibility tags](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/) for the wheels. The "python" tag indicates either the exact version supported by the wheel or the minimum supported version.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+Whether or not the version indicates a minimum or an exact version depends on the next ABI tag.
+The `py3` tag used by `pycparser` indicates that this is usably by _any_ Python 3 interpreter running _any_ Python version, not necesarily just CPython.
```
JelleZijlstra:
```suggestion
The `py3` tag used by `pycparser` indicates that this is usable by _any_ Python 3 interpreter running _any_ Python version, not necessarily just CPython.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+The `cp311` and `cp314` tag indicate that the `cryptography` and `cffi` wheels are only usable with CPython. Other Python implementations will need a different wheel.
+
+The ABI tag indicates the Python ABI the wheel file targets.
+The `none` ABI tag used by `pycparser` indicates that the wheel doesn't target a particular native ABI in particular: the wheel includes only pure-python code and binary compatibility doesn't need to be considered.
```
JelleZijlstra:
```suggestion
The `none` ABI tag used by `pycparser` indicates that the wheel doesn't target a particular native ABI in particular: the wheel includes only pure-Python code and binary compatibility doesn't need to be considered.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+This means that programming languages that cannot compile C syntax, like Rust, cannot use any C API items that are defined as C macros or inline functions.
+Instead, Rust extensions rely on Rust re-implementations of macros and static inline functions exposed by the C API based on items that _are_ in the Python ABI.
+C++ _is_ compatible with the C preprocessor, so you can use typedefs, macros, and inline functions exposed in the CPython C API in C++ extensions.
+
+### The PyObject struct
+
+In this section we do a bit of a deep dive on one of the structs that are part of the Python ABI.
+Why go into all this detail on this struct?
+For one, the `PyObject` struct is by far the most important struct in the CPython C API and the Python ABI.
+Also, because the layout of `PyObject` is one of the main sources of tension that led to the development of two new python ABIs in recent years.
```
JelleZijlstra:
```suggestion
Also, because the layout of `PyObject` is one of the main sources of tension that led to the development of two new Python ABIs in recent years.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+    Py_ssize_t ob_refcnt;
+    PyTypeObject *ob_type;
+}
+```
+
+Here, [`Py_ssize_t`](https://docs.python.org/3/c-api/intro.html#c.Py_ssize_t) is a signed integer type that is used to represent sizes in the CPython C API.
+In this case, the `ob_refcnt` field represents the [reference count](https://en.wikipedia.org/wiki/Reference_counting) of the object - the number of other data structures that reference an instance of the `PyObject` struct.
+This is used to manage memory: the reference count is incremented and decremented as an object is passed between different Python modules and functions.
+When the reference count goes to zero, the object is de-allocated.
+If you're curious why Python uses a signed integer to represent a strictly positive count: a signed integer would catostrophically wrap around to a large positive value if the reference count ever underflows.
```
JelleZijlstra:
```suggestion
If you're curious why Python uses a signed integer to represent a strictly positive count: a signed integer would catastrophically wrap around to a large positive value if the reference count ever underflows.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+With a signed integer you would see a negative reference count: an obviously invalid state.
+
+The other field, `ob_type` is a pointer to the type of the object.
```
JelleZijlstra:
```suggestion
The other field, `ob_type`, is a pointer to the type of the object.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+These days there is a much better distinction between CPython internals and publicly exposed API, with all signs pointing towards the trend of isolating interpreter internals to continue into the future.
+
+This is managed by breaking up the "full" C API surface used by the interpreter into five different layers, illustrated in the diagram below.
+
+ <figure style={{ textAlign: 'center' }}>
+   <img
+     src="/posts/python-abi-abi3t/cpython_api_layers.png"
+     alt='Five nested ellispes illustrating the layering of the Python C API. The outermost ellipse is gray and labeled "Internal API, exposed only if `Py_BUILD_CORE` is defined. The next enclosed ellipse is red and outlined with a dash line defined in the legend to mean "Usable with `#include "Python.h"`". The red ellipse is labeled "Private API `_Py*` prefix`. The next enclosed ellipse is yellow and is labeled "Unstable API `PyUnstable_*` prefix". The next enclosed ellipse is blue and labeled "Version-specific API". The next enclosed ellipse is green with a dark shaded outline the legend defines to mean "Usable if `Py_LIMITED_API` is defined and is labeled "Limited API".'
```
JelleZijlstra:
```suggestion
     alt='Five nested ellipses illustrating the layering of the Python C API. The outermost ellipse is gray and labeled "Internal API, exposed only if `Py_BUILD_CORE` is defined. The next enclosed ellipse is red and outlined with a dashed line defined in the legend to mean "Usable with `#include "Python.h"`". The red ellipse is labeled "Private API `_Py*` prefix`. The next enclosed ellipse is yellow and is labeled "Unstable API `PyUnstable_*` prefix". The next enclosed ellipse is blue and labeled "Version-specific API". The next enclosed ellipse is green with a dark shaded outline the legend defines to mean "Usable if `Py_LIMITED_API` is defined and is labeled "Limited API".'
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+     style={{position:'relative',left:'12%',width:'70%'}}
+   />
+ </figure>
+
+Below we describe each of these layers and explain how this separation enables many different usecases.
+
+### Internal C API
```
JelleZijlstra, line 203: The ordering of these sections feels strange, it goes back and forth between what feels more and less private.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+### Private C API
+
+For historical reasons, the public API includes many private implementation details whose names are prefixed by a leading underscore.
+To pick a random example: [`_PyDict_GetItem_KnownHash`](https://github.com/python/cpython/blob/2e5843e13fcfd768a435d82e6182af403844432c/Include/cpython/dictobject.h#L45-L46) allows users to bypass the normal Python `dict` interface to directly access a dict entry with an already-computed hash.
+This optimization is useful sometimes in the interpreter, so it's defined, but it's not a documented primitive, so it may either be deprecated and eventually removed or possibly get promoted to an official API function.
+Sometimes symbols are exposed like this for technical reasons, sometimes for historical reasons, and sometimes because a CPython user requested the ability to do so with the knowledge that there is no API stability guarantee for these functions.
+
+### Unstable C API
+
+The next innermost layer enclosees all of the documented functionality in the CPython C API.
```
JelleZijlstra:
```suggestion
The next innermost layer encloses all of the documented functionality in the CPython C API.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+### The Version-Specific API
+
+The next innermost layer is the Python version-specific C API and includes all symbols that follow the CPython C API [stability policy](https://docs.python.org/3/c-api/stable.html#c-api-stability), as originally adopted back in 2009 for [PEP 387](https://peps.python.org/pep-0387/).
+While public symbols in the CPython C API can be removed following a deprecation period or to fix a serious defect, this only happens after a period of public discussion and consideration of the community impact. Critically, CPython has a policy not to add or remove items in the version-specific API from the first beta release of a Python minor version onward.
+For example, this year Python 3.15.0b1 came out in May and that release froze the Python 3.15 version-specific C API.
+All subsequent betas, release candidates, and official releases share the same set of API items, including function singatures and type layours, with no changes allowed until the next Python minor release.
```
JelleZijlstra:
```suggestion
All subsequent betas, release candidates, and official releases share the same set of API items, including function signatures and type layouts, with no changes allowed until the next Python minor release.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Limited C API
+
+Finally, the most restricted innermost layer corresponds, fittingly, to the "Limited" C API.
+A somewhat confusing point that is often glossed over is that the limited C API _also_ changes from Python version to Python version. One can build extensions that target _either one_ of, say, the Python 3.13 version-specific API or the limited API.
+Either by defining a compiler macro or selecting an option in a build backend, one selects a particular limited API version to target.
+It's possible to target limited API versions older than the Python version used to build an extension.
+An extension or wheel built targeting, say, the Python 3.10 limited API can be imported on Python 3.10 and _any newer Python release_.
```
JelleZijlstra:
```suggestion
An extension or wheel build targeting, say, the Python 3.10 limited API can be imported on Python 3.10 and _any newer Python release_.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+While public symbols in the CPython C API can be removed following a deprecation period or to fix a serious defect, this only happens after a period of public discussion and consideration of the community impact. Critically, CPython has a policy not to add or remove items in the version-specific API from the first beta release of a Python minor version onward.
+For example, this year Python 3.15.0b1 came out in May and that release froze the Python 3.15 version-specific C API.
+All subsequent betas, release candidates, and official releases share the same set of API items, including function singatures and type layours, with no changes allowed until the next Python minor release.
+
+### Limited C API
+
+Finally, the most restricted innermost layer corresponds, fittingly, to the "Limited" C API.
+A somewhat confusing point that is often glossed over is that the limited C API _also_ changes from Python version to Python version. One can build extensions that target _either one_ of, say, the Python 3.13 version-specific API or the limited API.
```
JelleZijlstra: I feel like you're jumping into explaining what the limited API isn't before really explaining what it is.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+Of course "never" is a little strong, and if a particularly bad design flaw is discovered then the _implementation_ of the ABI entry corresponding to the pernicious symbol that is eventually removed from the limited API would probably trigger a runtime error.
+The _symbol_ would still be available in the ABI though, so an extension that happens to use the removed function would still compile, link, and import successfully, it just might trigger a runtime error when a new Python version uses functionality that _calls_ the removed function.
+
+These two ABI stability guarantees enable two different kinds of builds for distributors of extension modules:
+
+### The version-specific ABI: `cp3XY`
+
+The guarantee that the Python version-specific C API does not change within a Python minor release series corresponds to a stability policty for the CPython version-specific ABI.
+In practice, it means that extension modules and binary wheels built using Python 3.15.0b1 can be imported using any subsequent version of Python 3.15.
+Even years into the future.
```
JelleZijlstra:
```suggestion
In practice, it means that extension modules and binary wheels built using Python 3.15.0b1 can be imported using any subsequent version of Python 3.15, even years into the future.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+Projects shipping binary wheels using the version-specific ABI have ABI compatibility tags in the wheel filename like `cp314-cp314`. If a wheel has this particular tag, that indicates the wheel has binaries that are built using the GIL-enabled version-specific ABI for Python 3.14.
+We'll cover shortly why I had to add "GIL-enabled" there and why that's important.
+
+Because the version-specific ABI is particular to a particular minor release of CPython, this means a project must explicitly add support for new Python releases every year.
```
JelleZijlstra:
```suggestion
Because the version-specific ABI is specific to a particular minor release of CPython, this means a project must explicitly add support for new Python releases every year.
```
don't repeat "particular"

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+What is the GIL and why is it important for this discussion? It's actually very much related to the layout of the `PyObject` struct. As we saw above, `PyObject` contains a reference count field `ob_refcnt`, that holds the number of live referenes to the Python object.
+If the reference count goes to zero the object is deallocated.
+In CPython, the object is deallocated immediately.
+Critically, Python also exposes access to OS-level threads via the `threading` module and in principle, without a global lock preventing it, could expose _simultaneous_ access to the same Python object.
+
+So you have a situation where a C struct has a signed integer that may be incremented or decremented at arbitrary times -- even simultaneously -- from any arbitrary number of threads.
+This is a classic case of a problem that is susceptible to pitfalls of concurrency: [races](https://en.wikipedia.org/wiki/Race_condition).
+This particular case, where the race to increment or decrement the integer, happens in C code, is suseptible to a particularly bad kind of multithreaded race: a [data race](https://en.wikipedia.org/wiki/Race_condition#Data_race).
```
JelleZijlstra:
```suggestion
This particular case, where the race to increment or decrement the integer happens in C code, is susceptible to a particularly bad kind of multithreaded race: a [data race](https://en.wikipedia.org/wiki/Race_condition#Data_race).
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+In the specification of the C, C++, and Rust languages, the result of what happens after [these sorts of races are undefined behavior](https://www.hboehm.info/c++mm/why_undef.html).
+
+This problem was realized quite early on in the development of CPython.
+To fix it, the language grew the global interpreter lock.
+In order to interact with python objects, one needs to acquire a global [mutex](<https://en.wikipedia.org/wiki/Lock_(computer_science)>) and release it as soon as you are done using the python objects or the CPython interpreter runtime.
```
JelleZijlstra:
```suggestion
In order to interact with Python objects, you need to acquire a global [mutex](<https://en.wikipedia.org/wiki/Lock_(computer_science)>) and release it as soon as you are done using the Python objects or the CPython interpreter runtime.
```
The sentence uses both "one" and "you", pick one.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+Since incrementing and decrementing reference counts should only happen with the GIL mutex held, it's therefore perfectly safe to arbitrarily share Python objects: one can be confident that only one thread at a time can actually see or mutate the state of hte object.
+
+There is a big downside do this design decision: the GIL is a huge blocker for actually using more than one CPU core to run Python code.
```
JelleZijlstra:
```suggestion
There is a big downside to this design decision: the GIL means it is impossible to actually use more than one CPU core to run Python code.
```
Current wording is not terribly wrong but suggested wording feels clearer

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+In particular: only one core can run code that holds the GIL.
+Here, "code that holds the GIL" corresponds to all "pure" Python code that uses only builtins and does not use third-party code or the standard library.
+Some code in third-party extensions and the standard library does release the GIL and allow more than one CPU to simultaneously execute code, in practice one often finds that scaling is abysmal due to the impact of [Amdahl's "law"](https://en.wikipedia.org/wiki/Amdahl%27s_law): any code in an application that runs under one CPU can fundamentally not be parallelized and therefore blocks parallel scaling.
```
JelleZijlstra:
```suggestion
Some code in third-party extensions and the standard library does release the GIL and allow more than one CPU to simultaneously execute code, but in practice one often finds that scaling is abysmal due to the impact of [Amdahl's "law"](https://en.wikipedia.org/wiki/Amdahl%27s_law): any time-slice of an application's execution that must run on only one CPU will eventually prevent improving parallel scaling as a function of thread count.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,351 @@
...
+
+That is why there have been several efforts over the years -- to varying degrees of success -- to remove the GIL from the CPython implementation.
+Over the past few years, one of these approaches seems to have stuck.
+It looks like a free-threaded Python, one with no global interpreter lock and no limit to multithreaded concurrency in pure Python code, is the future of CPython
```
JelleZijlstra:
```suggestion
It looks like a free-threaded Python, one with no global interpreter lock and no limit to multithreaded concurrency in pure Python code, is the future of CPython.
```

JelleZijlstra (commented):
Found a number of typos and other small issues

ngoldbaum:
I think this still needs more visuals but I've gone ahead and marked it ready for review because I think the text is getting close to ready.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,543 @@
...
+What is the C API exactly?
+It's the set of C macros, typedefs, functions, and structs exposed by the header that defines the C interface: `Python.h`.
+It's possible to write C code that replicates the behavior of any Python script, at the cost of compilation, verbosity, and exposure to the pitfalls of the C programming language.
+The reward is raw execution speed.
+
+It is often possible to achieve order-of-magnitude or even several order-of-magnitude speedups by translating Python code to a compiled language that can call into the C API.
+The [Cython](https://cython.readthedocs.io/en/latest/) programming language takes advantage of this by (among other things) compiling Python code to C source that, once compiled and linked into a larger program, behaves like the Python it was generated from.
+This translation isn't always perfect, but it is serviceable enough to back the implementations of popular libraries like [Pandas](https://pandas.pydata.org/), [scikit-image](https://scikit-image.org/), or [scikit-learn](https://scikit-learn.org/).
+
+### The distinction between a C API and ABI
```
rgommers: For some mysterious reason, this section header doesn't render in the preview.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,543 @@
...
+
+What isn't the C API?
+The C API is purely a construct of the C programming language.
+Code written in languages that aren't C can call into a C API, but only by using the platform-specific and architecture-specific calling conventions used by all code that executes directly on the CPU, abstracting away functionality that is not expressible in C.
+This is when it becomes important to think about the ABI.
```
rgommers: This seems to imply (to me at least) that the ABI matters for C++, Rust et al., but not for C. That's not true of course, so you may want to tweak this just a little to avoid giving that impression.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,533 @@
...
+
+### What is a platform ABI?
+
+The platform ABI governs details of how exactly machine code executes on each architecture: how a compiler translates calling a C function like [`PyDict_GetItemRef`](https://docs.python.org/3/c-api/dict.html#c.PyDict_GetItemRef) into a concrete set of machine code that sets CPU registers and other platform-specific details per the specification of whatever platform the code ultimately runs on.
+
+A platform ABI determines things like the exact way machine code needs to pass arguments to a function. For example, on the [`x86_64`](https://wiki.osdev.org/System_V_ABI#x86-64) architecture, only the first six non-floating-point arguments of a function are passed via registers, the remaining arguments are passed via the stack.
+Normally, developers do not need to worry about details like this: compilers automatically generate machine code appropriate for whatever platform ABI a developer wants to target.
+However, it _is_ important to know that different operating systems and CPU architectures have unique ABIs and that code compiled for one platform ABI is completely unusable on another platform ABI.
```
rgommers: This paragraph is solid, so I'm not sure if it needs a change. However, it did occur to me here that both `libc` flavor and compiler toolchain choice go into the ABI, with practical consequences (e.g. manylinux vs. musllinux, or mingw-w64 incompatibility with MSVC). The phrase "libc" doesn't appear in the post at all currently.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,533 @@
...
+It's worth emphasizing that `abi3` wheels are _not_ installable on [the free-threaded build](https://py-free-threading.github.io) of CPython. We'll see below why, and how the new `abi3t` ABI in Python 3.15 fixes this.
+
+### The PyObject struct
+
+`PyObject` is by far the most important struct in the CPython C API and the Python ABI, and its layout is one of the main sources of tension that led to two new Python ABIs in recent years — so it's worth a closer look.
+
+All Python objects correspond in C to an instance of `PyObject` or a struct that extends the `PyObject` struct.
+Until free-threaded Python (more on that below), the `PyObject` struct had the following layout on 64-bit architectures:
+
+```C
```
rgommers: syntax highlighting is missing, here and further down - perhaps lowercase `c`?

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,533 @@
...
+  </table>
+</div>
+
+In the future, when Python 3.15 becomes your minimum supported Python version, you can go back to shipping one wheel per release.
+Temporarily building three wheels per release does add some ecosystem-wide cost with the reward of being able to adopt free-threaded Python.
+
+Currently, `abi3t` support in build tools and bindings generators is mixed at time of writing in June 2026, as Python 3.15 is still in a pre-release stage.
+As the Python 3.15 final release approaches, expect to see build backend support improved.
+[PyO3](https://pyo3.rs/v0.29.0/features.html#abi3t) and [Maturin](https://www.maturin.rs/bindings.html?highlight=abi3t#py_limited_apiabi3) fully support Python 3.15, so most Rust extensions should be ready for testing.
+[Scikit-build-core](https://scikit-build-core.readthedocs.io/en/latest/configuration/index.html#customizing-the-output-wheel) and [CMake](https://cmake.org/cmake/help/latest/module/FindPython3.html) also support abi3t.
```
rgommers: CMake isn't a build backend, I wouldn't mention it here. scikit-build-core relies on CMake having support in a pretty much identical way to meson-python relying on Meson.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,533 @@
...
+Cython supports abi3t via [an experimental branch](https://github.com/cython/cython/issues/7399).
+Setuptools support will be added once [PR #5193](https://github.com/pypa/setuptools/pull/5193) is merged and appears in a release.
+Meson-python support also lives in [an open PR](https://github.com/mesonbuild/meson-python/pull/856) currently.
```
rgommers: In case we release that before this post goes up (PR is about ready right now), let's update this:)

rgommers (commented):
This is looking very close now. I quite like the new figures! A few more comments.

apps/labs/posts/python-abi-abi3t.md
ngoldbaum: I was copy-editing while you were reviewing. I think this came in after I rearranged the text.

apps/labs/posts/python-abi-abi3t.md
ngoldbaum: I added a paragraph explaining that the above is a slight oversimplification and links to a nice blog post I found that goes into the complexities of target triples.

apps/labs/posts/python-abi-abi3t.md
ngoldbaum: I added a paragraph explaining how a C ABI in particular is very important for both C++ and Rust to enable binary interoperability.

apps/labs/posts/array-api-aot-jit.md
```diff
@@ -5,27 +5,25 @@ published: June 23, 2026
 description: 'In this blog post, we discuss ways of dealing with compiled code when adopting the Array API.'
 category: [Array API, GPU]
 featuredImage:
-  src: posts/array-api-aot-jit/array-api-aot-jit.png
+  src: /posts/array-api-aot-jit/array-api-aot-jit.png
   alt: 'Data APIs logo next to logos of NumPy, CuPy, PyTorch and JAX'
 hero:
-  imageSrc: posts/array-api-aot-jit/array-api-aot-jit.png
+  imageSrc: /posts/array-api-aot-jit/array-api-aot-jit.png
```
ngoldbaum: I was having trouble building the blog locally because of this.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,594 @@
...
+      </tr>
+      <tr>
+        <td>3.16</td>
+      </tr>
+      <tr>
+        <td>3.17 and later</td>
+      </tr>
+    </tbody>
+  </table>
+</div>
```
mgorny, line 635: Without visible grid, it's hard to tell which labels refer to which rows.

ngoldbaum: Do you have any idea how to fix it? This table was cargo-culted from another post.

ngoldbaum: Claude ended up coming up with something eventually that looks much nicer than what was there before. Thanks for the prompt.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,592 @@
...
+
+Below we describe each of these layers and explain how this separation enables many different use cases.
+
+### Internal C API
+
+The outermost layer is for people who work on the CPython implementation.
+In order to use it, a C or C++ program must define the `Py_BUILD_CORE` compiler macro, which is equivalent to declaring that your program is a part of the CPython interpreter itself.
+As [the CPython developer guide](https://devguide.python.org/developer-workflow/c-api/#the-internal-api) indicates, the internal API is subject to change at any moment and should not be relied on.
+Defining `Py_BUILD_CORE` and opting into using the internal C API is equivalent to saying you're willing to take on maintenance for code that can and will break at any time.
+For 99.9% of people who are not CPython contributors, the internal API should not be used.
```
encukou: Maybe it's worth mentioning that this API ties you to a specific *build* (rather than just *version*) of CPython -- so it has some use for people who build a modified interpreter from source.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,592 @@
...
+Normally, developers do not need to worry about details like this: compilers automatically generate machine code appropriate for whatever platform ABI a developer wants to target.
+However, it _is_ important to know that different operating systems and CPU architectures have unique ABIs and that code compiled for one platform ABI is completely unusable on another platform ABI.
+This fact determines much of the design of the [binary wheel distribution format](https://packaging.python.org/en/latest/specifications/binary-distribution-format/), which we'll discuss in more detail below.
+
+The biggest consequence is that each platform and CPU architecture, each with its own distinct platform ABI, requires its own unique builds.
+This is one reason projects like `NumPy` distribute so many binary wheels with each release: projects need to build binaries for each platform ABI they want to support.
```
encukou: Maybe link "so many binary wheels" to https://pypi.org/project/numpy/2.5.0/#files ?

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,592 @@
...
+
+### Private C API
+
+This layer is _private_ in the sense that its items shouldn't be used outside CPython, but "public" in the sense that they're available after a plain `#include "Python.h"`, with no special macros.
+Names in the C API with a leading underscore are private.
+To pick a random example: [`_PyDict_GetItem_KnownHash`](https://github.com/python/cpython/blob/2e5843e13fcfd768a435d82e6182af403844432c/Include/cpython/dictobject.h#L45-L46) allows users to bypass the normal Python `dict` interface to directly access a dict entry with an already-computed hash.
+This optimization is useful sometimes in the interpreter, so it's defined.
+It's not a documented primitive, so there are no guarantees about it being available in a future version of the C API.
+It _is_ part of the Python ABI, though, so it inherits the ABI's stability guarantees even though it carries no API guarantees.
+Sometimes symbols are exposed like this for technical reasons, sometimes for historical reasons, and sometimes because a CPython user requested the ability to do something and was OK with the lack of API stability.
```
encukou: I'd omit the "because a CPython user requested" -- that's nowadays all "historical reasons". Or *mistakes* (not every core dev knows how this should work, and there's no police).

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,592 @@
...
+
+### The Version-Specific C API
+
+The next innermost layer is the Python version-specific C API and includes all symbols that follow the CPython C API [stability policy](https://docs.python.org/3/c-api/stable.html#c-api-stability), as first laid out in [PEP 387](https://peps.python.org/pep-0387/) back in 2009.
+While public symbols in the CPython C API can be removed following a deprecation period or to fix a serious defect, this only happens after a period of public discussion and consideration of the community impact. Critically, CPython has a policy not to add or remove items in the version-specific API from the first beta release of a Python minor version onward.
+For example, this year Python 3.15.0b1 came out in May and that release froze the Python 3.15 version-specific C API.
+All subsequent betas, release candidates, and official releases share the same set of API items, including function signatures and struct layouts, with no changes allowed until the next Python minor release.
+This also means the ABI is frozen.
```
encukou: This does not follow -- the ABI chan change independently of the ABI.

```suggestion
The ABI is also frozen when the first release candidate of a new minor version is released.
```

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,592 @@
...
+However, it's not very hard to find real-world examples of this pattern.
+For example, [in NumPy](https://github.com/numpy/numpy/blob/4357ad1a67e7d4a686c39f3720604b6440725e40/numpy/_core/include/numpy/arrayscalars.h#L140-L147).
+
+This pattern only works if the layout of e.g. `PyDictObject` is known at compile time. However, if `PyObject` is opaque, then `PyDictObject` must also be opaque, and it's no longer possible to define C subtypes by simply defining a struct with a `base` member.
+Instead, you need to use [a new API specifically for this case](https://peps.python.org/pep-0697/), added in Python 3.12.
+
+As of 2025, during the Python 3.15 development cycle, there were still a few more spots that needed updating:
+
+- [A new C API](https://peps.python.org/pep-0793/) for defining modules that avoids `PyModuleDef`, a type that extends `PyObject`.
+- [An accompanying new API](https://peps.python.org/pep-0820/) for defining Python types and modules using a common `PySlot` struct.
```
encukou: PEP 820 wasn't *needed*; I think this post can safely omit it. (PEP 793 [should get](https://github.com/python/peps/pull/5006) notes about the API change, and of course the [docs](https://docs.python.org/3.15/c-api/extension-modules.html#extension-modules) don't mention PEP-793-without-820 at all).

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,592 @@
...
+category: [PyData ecosystem]
+featuredImage:
+  src: /posts/python-abi-abi3t/cpython_api_layers_listing.png
+  alt: 'Five nested ellipses illustrating the layering of the Python C API. The outermost ellipse is gray and labeled "Internal API". The next enclosed ellipse is red and is labeled "Private API". The next enclosed ellipse is yellow and is labeled "Unstable API". The next enclosed ellipse is blue and labeled "Version-specific API". The next enclosed ellipse is green and is labeled "Limited API".'
+hero:
+  imageSrc: /posts/python-abi-abi3t/cpython_api_layers_hero.png
+  imageAlt: 'Five nested ellipses illustrating the layering of the Python C API. The outermost ellipse is gray and labeled "Internal API". The next enclosed ellipse is red and is labeled "Private API". The next enclosed ellipse is yellow and is labeled "Unstable API". The next enclosed ellipse is blue and labeled "Version-specific API". The next enclosed ellipse is green and is labeled "Limited API".'
+---
+
+# What Every Python Developer Should Know About the CPython ABI
```
encukou, line 15: The title might be too clickbaity? You shouldn't need to care until you want to *build* extension modules...

apps/labs/posts/python-abi-abi3t.md
encukou: I've been [using](https://docs.python.org/3/c-api/stable.html#platform-considerations) the term “the (underlying) platform” for this. Maybe you can too?
```suggestion
However, it _is_ important to know that the underlying platform -- details like
different operating systems and CPU architectures -- have unique ABIs and that code compiled for one platform ABI is completely unusable on another platform ABI.
```

encukou (commented):
A great blog post!
I do have a few comments, but nothing major:

apps/labs/posts/python-abi-abi3t.md
ngoldbaum: Yeah, it’s definitely clickbait. This is the problem with needing to come up with talk titles before you actually write the talk. I’ll think about whether the blog post should have a different title but the ship has probably sailed on the title for the EuroPython talk.

apps/labs/posts/python-abi-abi3t.md
```diff
@@ -0,0 +1,601 @@
...
+     style={{position:'relative'}}
+   />
+ </figure>
+
+Below we describe each of these layers and explain how this separation enables many different use cases.
+
+### Internal C API
+
+The outermost layer is for people who work on the CPython implementation.
+In order to use it, a C or C++ program must define the `Py_BUILD_CORE` compiler macro, which is equivalent to declaring that your program is a part of the CPython interpreter itself.
```
kumaraditya303, line 206: `Py_BUILD_CORE` is actually not enough, the symbol should also be visible which isn't the case for many of such internal APIs, maybe this could be a note?

rgommers:
Looks great, in it went - thanks Nathan and all reviewers!

PR #991, opened 2026-08-17
Fix rendering issue in the abi3t blog post

ngoldbaum:
The LLM-generated `dangerouslySetInnerHTML` hack in my post seems to be broken. Let's delete it and use mdashes instead.


rgommers (approved):
Thanks Nathan

PR #992, opened 2026-08-19
add generated categories/* links (issue 988)

### LINKED ISSUE #988: [BUG] - Unable to link to blog categories
### What site is this for?

Quansight Labs

### Expected behavior

Links to specific blog categories should work, like: https://labs.quansight.org/categories/documentation/

This is useful and is likely breaking some links within blogs posts.

### Actual behavior

You can filter by categories within https://labs.quansight.org/blog, but it does not create specific URLs for each category.

https://labs.quansight.org/categories/documentation/ leads to 404

### Anything else?

Originally noted by @melissawm

mattip:
Fixes #988 by adding a categories route

AI was used to program this, I tested it locally.

### DIFF
--- a/apps/labs/src/pages/categories/[category].astro
+++ b/apps/labs/src/pages/categories/[category].astro
@@ -1,0 +1,38 @@
+---
+import { getEntry } from 'astro:content';
+
+import BlockRenderer from '../../components/BlockRenderer.astro';
+import { BlogList } from '../../components/Blog/BlogList';
+import BaseLayout from '../../layouts/BaseLayout.astro';
+import categories from '../../../posts/categories.json';
+import { slugifyCategory } from '../../utils/categories';
+import { getPublishedPosts } from '../../utils/posts';
+
+export function getStaticPaths() {
+  return categories.map((category) => ({
+    params: { category: slugifyCategory(category) },
+    props: { category },
+  }));
+}
+
+const { category } = Astro.props;
+const page = await getEntry('pages', 'blog');
+const posts = await getPublishedPosts();
+---
+
+<BaseLayout
+  title={`${category} | Blog | Labs`}
+  description={page?.data.description}
+>
+  {page?.data.blocks.map((block) => (
+    <BlockRenderer block={block} />
+  ))}
+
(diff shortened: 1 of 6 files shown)


mattip:
works as advertised, for instance this preview link https://labs-git-categories-quansight.vercel.app/categories/gpu shows all the gpu-tagged posts

apps/labs/src/pages/categories/[category].astro
```diff
@@ -0,0 +1,38 @@
...
+<BaseLayout
+  title={`${category} | Blog | Labs`}
+  description={page?.data.description}
+>
+  {page?.data.blocks.map((block) => (
+    <BlockRenderer block={block} />
+  ))}
+
+  <div class="pt-[3rem] mx-auto w-[95%] max-w-[83rem] md:w-[85%] xl:w-[70%]">
+    <h2 class="text-[2.4rem] font-extrabold leading-[4.9rem] text-violet font-heading">
```
Comment (trallard) [inline apps/labs/src/pages/categories/[category].astro:32]: Don't we already have reusable styles/classes defined? This PR seems to have a few instances of inline styling. 
It is ok for now if this is more of a quick fix while we continue with the redesign.

trallard (approved):
Left a comment (more out of cuuriosity than anything else).
This seems to work and I could not spot any obvious issues.

apps/labs/src/pages/categories/[category].astro
Comment (mattip) [reply apps/labs/src/pages/categories/[category].astro:32]: I used AI to do the change, it is not very good about compact CSS classes. I can ask it to refactor. @pavithraes thoughts?

Comment (trallard) [reply apps/labs/src/pages/categories/[category].astro:32]: Makes sense. Let's leave it as it is since we have the whole redesign planned and I do hope we won't need to make many more changes between now and then.

Comment (trallard) [reply apps/labs/src/pages/categories/[category].astro:32]: @pavithraes unless you have an objection I think we can approve and merge this?

Comment (pavithraes) [reply apps/labs/src/pages/categories/[category].astro:32]: I agree that we can leave it as-is, considering the redesign coming soon. Merging sounds good!

Issue #982, opened 2026-07-30
[BUG] - social previews are hitting a 429 error


ngoldbaum:
### What site is this for?

Quansight Labs

### Expected behavior

I would expect discourse to be able to generate a social preview.

### Actual behavior

<img width="906" height="714" alt="Image" src="https://github.com/user-attachments/assets/cca5d13a-b1ab-44df-91dc-be3507f90b66" />

### How to Reproduce the problem?

Link to a quansight labs blog post in a context that generates a social preview.

mattip:
#980 changed the hosting, I think now you get a link without the error message.

mattip:
#994 was merged. Did the change propogate through the caches?

ngoldbaum:
Yeah this has been fixed for a while thanks.

PR #980, opened 2026-06-09
Migrate off storyblok

mattip:
This is the first part of a migration task to remove the integration with storyblock and use astro to render the blocks. I used claude to help me download all the data we currently have in storyblok. The details are in the storyblok-migration-download.md document, and the scripts are part of the PR. The bulk of the data is the 106 team members, and then there are 7 pages, and the header and footer for the site.

PR #994, opened 2026-08-20
add social media tags

mattip:
Related to #982, which changed when the site moved away from storyblock in #980. Add social media tags so pasting a blog post into social media will show an image and a blurb.

### DIFF
--- a/apps/labs/astro.config.mjs
+++ b/apps/labs/astro.config.mjs
@@ -4,7 +4,20 @@ import react from '@astrojs/react';
 import rehypeKatex from 'rehype-katex';
 import remarkMath from 'remark-math';
 
+// `site` is what BaseLayout builds absolute og:url/og:image from. On a
+// Vercel preview build, point it at that deployment so the tags describe
+// the preview being tested rather than production; VERCEL_BRANCH_URL is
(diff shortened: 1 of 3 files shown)


trallard (commented):
Thanks @mattip! This seems to be working as expected. 
However, since it seems we are using the ["dark theme" version of the Labs logo](https://github.com/Quansight/Quansight-website/blob/main/apps/labs/public/images/logos/labs-logo-footer-desktop.png) this is not visible in some instances, see:

(light theme, also not visible in Mastodon, Bluesky, and threads)
Blog___Labs_-_Polypane

(dark theme, also not visible in Mastodon)
Blog___Labs_-_Polypane

I am not sure if it is worth changing this for another version of the logo or leaving it as is for now since we are doing the whole website overhaul. From an accessibility perspective, it makes sense to change the image, but I'm not sure if we have one that would work out of the box for all scenarios.

On the other hand,  the blog image previews seem to work as expected
Notification_Center

mattip:
Easy enough to change, how about this one? I committed the change, but I am not sure the media links update, they do some heavy caching. blog_feature_var2

trallard (approved):
Ok let's ship this with the new social card image 🚀

PR #995, opened 2026-08-26
Add "Teaching NumPy's ufuncs new tricks" internship blog post

ikrommyd:
## Text styling

## Non-text contents


rgommers:
I had a quick browse, looks like a nice post! The one immediate question about the way this post is structured is that the end is quite abrupt - maybe a conclusion or "what may come next" to wrap it up would help?

@ngoldbaum I assume you'll review in detail?

ikrommyd:
Indeed, I will write an actual ending. Thanks!

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+I ended up doing a PhD in High-Energy Physics, a field that collects petabytes of data per experiment per year, as a member of the [CMS](https://cms.cern) experiment at [CERN](https://home.cern), where I worked heavily with the [Scikit-HEP](https://scikit-hep.org) stack of tools.
+I wanted to program more and I was also learning more about open source software, so I started cold-messaging people like "Hey, you maintain this tool, I'm a HEP student, I want to work on it", and I ended up becoming a maintainer of a few of those tools, most notably [Awkward Array](https://awkward-array.org) and [Coffea](https://coffea-hep.readthedocs.io).
+Scikit-HEP is itself a member of [Scientific Python](https://scientific-python.org), so I gradually got more involved there too, while my love for physics was going down and my love for programming and scientific computing was going up.
+At [SciPy 2025](https://www.scipy2025.scipy.org) I hacked a bit on NumPy during the sprints and met [Nathan Goldbaum](https://github.com/ngoldbaum), who would become my mentor and who told me to apply for the Quansight internship program, so I did, and I was given the opportunity to officially work on NumPy!
```
ngoldbaum:
```suggestion
At [SciPy 2025](https://www.scipy2025.scipy.org) I hacked a bit on NumPy during the sprints and met [Nathan Goldbaum](https://github.com/ngoldbaum), who would become my mentor and who told me to apply for the Quansight internship program.
A year later, I was accepted into the 2026 Quansight internship cohort where I had the opportunity to work on NumPy!
```

Take it or leave it but I think it reads clearer like this.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+My first task was to implement `np.minmax` so let's jump right on that and see why that was not trivial in NumPy before.
```
ngoldbaum: I think you can probably delete this sentence since the next paragraph becomes a rehash of this

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+Perhaps this paragraph did not make a lot of sense and there were some unknown words but everything will be explained in the rest of this post. To do so, we have to start by explaining how some NumPy internals work.
+
+### What are NumPy's universal functions
+
+NumPy has this concept of "universal functions" or "[ufuncs](https://numpy.org/doc/stable/user/basics.ufuncs.html)" for short. Ufuncs are functions that operate on arrays in an element-wise fashion. They are "vectorized" and they take in a fixed number of inputs and produce a fixed number of outputs. They also support broadcasting and typecasting and other standard features. For example [numpy.add](https://numpy.org/doc/stable/reference/generated/numpy.add.html#numpy.add) is a ufunc that takes in two input arrays and produces one output array whose elements are the result of the element-by-element addition of the elements of the two input arrays.
```
ngoldbaum:
```suggestion
NumPy has a concept of "universal functions" or "[ufuncs](https://numpy.org/doc/stable/user/basics.ufuncs.html)" for short.
Ufuncs are functions that operate on arrays in an element-wise fashion. They are "vectorized" and they take in a fixed number of inputs and produce a fixed number of outputs.
They also support broadcasting and typecasting and other standard features.
For example, [numpy.add](https://numpy.org/doc/stable/reference/generated/numpy.add.html#numpy.add) is a ufunc that takes two arrays as input arguments and writes to one output array whose elements are the result of the element-by-element addition of the elements of the two input arrays.
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+>>> np.add(x1, x2)
+array([[ 6,  8],
+       [10, 12]])
+```
+
+and as we mentioned earlier broadcasting and typecasting is supported
+
+```python
+>>> integer_array = np.array([[1,2],[3,4]])
+>>> float_array = np.array([1.1,2.2])
```
ngoldbaum: For readers who aren't aware of NumPy's conversion rules IMO it's clearer to explicitly say what the dtype is.


```suggestion
>>> integer_array = np.array([[1,2],[3,4]], dtype=np.int64)
>>> float_array = np.array([1.1,2.2], dtype=np.float64)
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+>>> np.add(integer_array, float_array)
+array([[2.1, 4.2],
+       [4.1, 6.2]])
+```
+
+where you can notice that while `x1` is an array of integers and is 2x2, `x2` is an array of doubles and is only a single row of two elements so integers need to be upcasted to doubles to do the addition and also `x2` is broadcasted to every row of `x1` to make the shapes match and do the addition.
```
ngoldbaum: This paragraph still refers to `x1` and `x2`. It's also a little confusing. Try to break it up into multiple sentences.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+There are also [generalized ufuncs](https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html#c-api-generalized-ufuncs) which are functions over vectors (or arrays) of elements.
```
ngoldbaum:
```suggestion
There are also [generalized ufuncs](https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html#c-api-generalized-ufuncs) which are functions over arrays of elements.
```

The distinction between an array and a vector isn't worth distracting the reader over IMO

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+They effectively operate on a subarray-by-subarray basis instead of element-by-element. We'll talk more about those later.
```
ngoldbaum:
```suggestion
They operate on a subarray-by-subarray basis instead of element-by-element.
We'll talk more about those later.
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+Everything else in this post is a variation on these loops, so it is worth going through the anatomy of one before we go any further.
+Reductions, the multi-output reductions we will get to later, and generalized ufuncs are all differences in what a loop is handed and how NumPy calls it, and none of that will make sense without seeing the plain element-wise case first.
```
ngoldbaum:
```suggestion
Reductions and generalized ufuncs are all implemented in NumPy in terms of what a loop is handed and how NumPy calls it.
However, none of that will make sense without seeing the plain element-wise case first.
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+So let's write one, piece by piece. This is an illustrative sketch of one of the `numpy.add` loops, the one for the `double` data type.
+
+Every ufunc loop has the same signature:
+
+```c
+void double_add_ufunc_loop(char **args, const npy_intp *dimensions,
+                           const npy_intp *steps, void *data)
+```
+
+NumPy does not hand the loop arrays, it hands it raw memory plus instructions on how to walk it.
```
ngoldbaum:
```suggestion
NumPy does not hand the loop arrays, it hands it pointers to memory a buffer alongside instructions on how to iterate over the buffer.
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+`args` holds a pointer to the start of each operand's data buffer (the underlying data buffers of NumPy arrays), `dimensions` says how many elements to process, and `steps` says how to get from one element to the next in each operand.
```
ngoldbaum:
```suggestion
The `args` holds a pointer to the start of each operand's data buffer (the underlying data buffers of NumPy arrays), `dimensions` holds the number of elements along each dimension of the array, and `steps` says how to get from one element to the next in each operand.
```

I don't like starting sentences with lower-case variable names, just a personal style thing.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+`data` is optional extra state a loop can be registered with, which we will not use here.
```
ngoldbaum:
```suggestion
The `data` is optional extra state a loop can be registered with, which we will not use here.
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+The first thing to read out is how many elements this call has to process:
```
ngoldbaum:
```suggestion
The first thing we need to do to implement a loop is read out is how many elements this call has to process:
```

otherwise it's not clear what the antecedent to  "the first thing" is.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+```c
+npy_intp n = dimensions[0];
+```
+
+Note that this is not necessarily the size of the array. NumPy may split the work into chunks and call the loop several times, so `n` is just how much work this particular call was given.
```
ngoldbaum, line 129: You might also point out that this simplified example is for 1D arrays.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+    in1_ptr += in1_step;
+    in2_ptr += in2_step;
+    out_ptr += out_step;
+}
+```
+
+Each iteration casts the two input pointers to `double *` and dereferences them to get the current pair of values, adds them, writes the result through the output pointer, and then advances all three pointers by their strides to land on the next element.
+
+Putting it all together, that is the whole loop:
```
ngoldbaum: I would personally show the full loop then break it out afterwards. Otherwise you don't have any context what the explanation above is about.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+array([3, 7])
+>>> np.sum([[1,2],[3,4]], axis=None)
+np.int64(10)
+>>> np.sum([[1,2],[3,4]], axis=(0, 1))
+np.int64(10)
+```
+
+In the examples above, a 1-dimensional array is reduced down to a scalar, and a 2-dimensional array is reduced down to either a 1-dimensional array if one dimension is collapsed or a scalar if both are collapsed. But how does NumPy actually perform those operations?
+
+One might think that there are specialized compiled kernels/loops (like the ufunc loops) to do reductions. What if I told you that all such reductions in NumPy are actually using regular forward ufunc loops underneath (like the one we wrote earlier)? But how is this possible? Reductions are a wrapper a method of the ufunc called `reduce`, so `numpy.sum` is equivalent to `numpy.add.reduce`:
```
ngoldbaum:
```suggestion
One might think that there are specialized compiled kernels/loops (like the ufunc loops) to do reductions. It turns out all simple reductions in NumPy are actually using regular forward ufunc loops underneath, like the one we wrote earlier.
But how is this possible?
Reductions are a wrapper a method of the ufunc called `reduce`, so `numpy.sum` is equivalent to `numpy.add.reduce`:
```

"What if I told you" makes me think of a speech Morpheus gives to Neo in The Matrix 😂. Take it or leave it.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+```python
+>>> np.add.reduce([1,2,3])
+np.int64(6)
+```
+
+To do a summation, NumPy is actually using the `numpy.add` ufunc and its loops. Similarly, `numpy.prod` is using the `numpy.multiply` ufunc, `numpy.min` is using the `numpy.minimum` ufunc, and so on.
```
ngoldbaum:
```suggestion
Similarly, `numpy.prod` uses the `numpy.multiply` ufunc, `numpy.min` is using the `numpy.minimum` ufunc, and so on.
```

The first sentence repeats the previous discussion. There's a lot of passive voice use in this blog post - try to give it a pass to fix that, especially if it allows you to write a little more efficiently like this.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+But how can the loop we wrote in C earlier be used to do a summation? It isn't immediately obvious. Let's consider the following. If we want to sum a 1-dimensional array like `[1,2,3]` to get 6, we need an accumulator that we keep adding all the values of the array to. That accumulator usually starts at 0. Then we need to iterate over the array, adding the current element on the accumulator
```
ngoldbaum:
```suggestion
But how can the loop we wrote in C earlier be used to do a summation? 
Let's consider summing a 1-dimensional array like `[1,2,3]` to get 6.
We need an accumulator to store the result.
For integers, the default initial value for reductions is 0, but you can also choose a different value.
Then we need to iterate over the array, adding elements and keeping a running sum in the accumulator.
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+>>> arr = [1,2,3]
+>>> accum = 0
+>>> for i in arr:
+...     accum += i
+...
+>>> accum
+6
+```
+
+Can the `double_add_ufunc_loop` loop be used to do that? Well if we make the `args[0]` input pointer point to the accumulator location, and we set `args[2]` to be equal to `args[0]` (the output location is the same as the first input pointer location and both are therefore the accumulator location) and we also set `steps[0] = steps[2] = 0` (the accumulator does not move) then if we also seed the accumulator with the value 0, the loop does what we want.
```
ngoldbaum:
```suggestion
Can the `double_add_ufunc_loop` loop be used to do that? Well if we set the `args[0]` input pointer to the accumulator location, and we set `args[2]` to be equal to `args[0]` (the output location is the same as the first input pointer location and both are therefore the accumulator location) and we also set `steps[0] = steps[2] = 0` (the accumulator does not move) then if we also seed the accumulator with the value 0, the loop does what we want.
```

This sentence is pretty long and complicated. Consider breaking it up.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+And this is what NumPy does in the `reduce` method of ufuncs. To do summation in particular, it steers the `numpy.add` loop effectively like this to reduce over an array whose data buffer is `data` and has `n` elements and a stride `stride` to get the next element
```
ngoldbaum:
```suggestion
And this is what NumPy does in the `reduce` method of ufuncs. To do summation in particular, it steers the `numpy.add` loop like this to reduce over an array whose data buffer is `data` and has `n` elements and a stride `stride` to get the next element
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+ValueError: reduce only supported for functions returning a single value
+>>> np.frexp.reduce([1,2,3])
+Traceback (most recent call last):
+  File "<python-input-2>", line 1, in <module>
+    np.frexp.reduce([1,2,3])
+    ~~~~~~~~~~~~~~~^^^^^^^^^
+ValueError: reduce only supported for binary functions
+```
+
+`reduce` was only supported for ufuncs that are 2-in/1-out and that sort of made sense from a structural perspective.
```
ngoldbaum:
```suggestion
The `reduce` operation is only supported for ufuncs that are 2-in/1-out.
```

Not sure what you're saying in the second sentence, maybe reword instead of deleting.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+First of all reducing means that there's some sort of comparison between two arrays so the two input limitation is kind of intuitive and the one output limitation is due to the fact that the reductions were always using the forward ufunc loops to perform them.
```
ngoldbaum:
```suggestion
Reducing means that there's some sort of comparison between two arrays, so the two input limitation is intuitive in that sense. The one output limitation is due to the fact that the reductions were implemented by re-using the forward ufunc loops.
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+A reduction that accumulates N values, when we think about it at the low level, needs N accumulation locations (pointers) and one more pointer that slides over the reduced array. So if a ufunc is `2-in/N-out`, it means that the loop uses two input pointers and N output pointers.
+There aren't enough pointers to do the trick where we'd set N input pointers equal to N output pointers equal to the N accumulator locations in a similar way we did for the `numpy.add` loop (where N = 1).
+Reductions using such a trick can be performed only when the loop is already `N + 1-in/N-out`.
```
ngoldbaum: This paragraph is re-stating what you're saying in the previous paragraph, but more technically. IMO you could just delete it. If you want to keep the detail, I'd just explain it once and integrate the low-level detail into the first paragraph.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+### np.minmax
+
+A long-requested feature in NumPy has been a `numpy.minmax` reduction that returns the minimum and maximum value in a single pass. Of course that is structurally equivalent to
```
ngoldbaum:
```suggestion
A long-requested feature in NumPy has been a `numpy.minmax` reduction that returns the minimum and maximum value in a single pass.
That is structurally equivalent to
```

Maybe also link to the old feature request issue asking for minmax?

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+```python
+def minmax(x, ...):
+    return numpy.min(x, ...), numpy.max(x, ...)
+```
+
+but not performance-wise.
+
+That is a reduction with two outputs. `numpy.min` uses the `numpy.minimum` loop that returns the element-wise minima of two arrays and `numpy.max` uses the `numpy.maximum` loop that returns the element-wise maxima of two arrays.
+If we implemented a ufunc like `numpy.minimummaximum` that returns the element-wise minima and maxima of two arrays in a single pass, we still couldn't implement `numpy.minmax` because the `reduce` method of such a ufunc wouldn't work, since it's a 2-in/2-out ufunc as explained in the previous section.
```
ngoldbaum:
```suggestion
If we implemented a `numpy.minimummaximum` ufunc that returns the element-wise minima and maxima of two arrays in a single pass, we still couldn't implement `numpy.minmax` because the `reduce` method of such a ufunc wouldn't work, since it's a 2-input/2-output ufunc.
```

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+So, on top of its forward element-wise loop, a ufunc's loop implementation can now register an optional, dedicated _reduction loop_, and `reduce` will pick that one up and steer it instead of the forward one.
+
+The first question is how many inputs and outputs such a loop should have.
+Going back to the low level picture from the previous section, a reduction that accumulates N values needs the N current accumulators coming in, plus the one element that is currently being streamed in from the array being reduced, and it produces the N updated accumulators.
+That is an `N + 1-in/N-out` loop.
+The nice thing about this convention is that for `N = 1` it collapses to `2-in/1-out`, which is exactly the shape of the classic forward loops that `reduce` has always been steering.
+In other words, the way `numpy.add.reduce` has always worked is just the `N = 1` special case of the new signature, so the machinery needs only one code path and nothing had to change for the existing single-output ufuncs.
+
+Concretely, such a loop is handed its data pointers and its strides in this order
```
ngoldbaum, line 340: This is a little abstract, it might help to expand this into a code example that explains what "is handed" means exactly

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+
+    /* Run the reduction loop we wrote above */
+    double_minimummaximum_reduce_loop(args, dimensions, steps, NULL);
+
+    /* The loop left the results sitting in the two accumulators */
+    return (minmax_t){ accum_min, accum_max };
+}
+
+double arr[5] = {3.0, 1.0, 4.0, 1.0, 5.0};
+minmax_t res = minmax_reduce(arr, 5, sizeof(double));   /* res.min = 1.0, res.max = 5.0 */
```
ngoldbaum, line 494: It's a little weird having this example that uses the C API you created directly, it's the only place it happens in the whole post (I think). It might be a little clearer if you broke it out into its own example and added some text to explain what you're trying to communicate about how the function works and how the result is stored in a struct.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+(array([0, 3, 6]), array([2, 5, 8]))
+```
+
+`numpy.minmax` landed in [numpy/numpy#32231](https://github.com/numpy/numpy/pull/32231).
+
+### Segmented reductions
+
+A segmented reduction is a reduction applied to consecutive chunks of an array instead of to the whole thing.
+It is the natural operation on ragged data: if you have a bunch of variable-length lists flattened into a single array, plus an array of offsets telling you where each list starts and ends, then "the sum of each list" is a segmented sum.
+That is how [Awkward Array](https://awkward-array.org) stores its data, it is also what the rows of a CSR sparse matrix look like, and it is why pretty much every array library that touches this kind of data ends up with some version of the operation.
```
ngoldbaum: I'd add a link for CSR sparse matrix too

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+        in_ptr += in_step;
+        out_ptr += out_step;
+    }
+}
+```
+
+That inner loop is the whole point: the kernel sees an entire subarray at once, so operations that are not element-wise at all can still be written as a single pass in C, and they get broadcasting, dtype resolution, `out` and subclass handling from the ufunc machinery for free.
+The [generalized ufunc API docs](https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html#c-api-generalized-ufuncs) have the full rules for signatures and for what the loop is handed.
+
+The first function I converted was `numpy.unwrap`, and there the motivation was performance.
```
ngoldbaum: Maybe link to the issue about this? And explain that the existing implementation was written in Python, which is simple but loses performance and features.

ngoldbaum: Also link to the numpy docs to explain what `numpy.unwrap` does, along with a one-sentence explanation. It's a little obscure.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,636 @@
...
+After that, it is worth exploring what else the new reduction loops make possible. Anything that can be expressed as a reduction operation can now get its own dedicated single-pass loop, and computing the mean and the variance of an array in a single pass is one example that comes to mind.
+
+On a personal note, this internship was a very valuable learning experience.
+I came into it with essentially no C experience: I had read [_The C Programming Language_ (K&R)](https://en.wikipedia.org/wiki/The_C_Programming_Language) and the [Extending and Embedding the Python Interpreter](https://docs.python.org/3/extending/index.html) docs, and built a [small C extension](https://github.com/ikrommyd/tensor.c) to learn how C extensions are written, and that was about it, so getting to learn NumPy's internals was more valuable to me than I can express.
```
ngoldbaum: Now go get a copy of effective C: https://nostarch.com/effective-c-2nd-edition 😄

ngoldbaum (commented):
Nice!

The intro reads great besides my small nits. I know we had a few review rounds there but I think it's the perfect length now and doesn't distract from the rest of the content.

I have lots of mostly nitpick copy-editing comments. However, do give this a pass (maybe with an LLM, but don't let it directly edit the post, instead take, leave, or edit suggestions) to find complicated multi-clause sentences that can be broken up or rearranged. Also do a pass for passive voice. Sometimes passive voice is ok but you can often make sentences easier to read without it.

This is very code-heavy. There's also no images besides the numpy logo and your headshot. I think in a few spots it might be nice to also have some diagrams for visual learners. Take or leave any of these suggestions:

* The structure of a strided loop, with arrows colored rows of squares to represent the input and output arrays.
* An illustration of how a reduction works, with colored boxes to represent the accumulator and array in 1D and then a generalization to 2D with an array of accumulators.
* Maybe a diagram that somehow visually shows why multi-output reductions don't fit into the above diagram?
* A diagram showing how a dedicated reduction loop resolves this conflict

ikrommyd:
Thanks for the thorough review @ngoldbaum! It's a bit late for me right now so I will deal with this tomorrow. Regarding the diagrams, I'm not particularly good at drawing such things but I'll see what I can do. I've done like 1-2 in the past by making svgs with AI.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
ikrommyd: image

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
ikrommyd: Even for 1D arrays I think the loop can be called multiple times in batches when casting is involved for example. Or do you mean something else?

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
ikrommyd: I do the same earlier where I do `add_reduce` in C to explain what NumPy effectively does to call the loop. This is similar.

ikrommyd: I personally like this idea to simply show in C how numpy would steer and call the loop. Would help me understand it as a reader and since the whole post assumes being able to read basic C I think it's fine.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
ikrommyd: It's explained a lot better at the `minmax_reduce` snippet below. I'll reference that and also added a sentence explaining how this directly collapses down to what the `add_reduce` snippet earlier does.

ikrommyd:
I think I've addressed most of the review comments. I've left as unresolved the ones I didn't do any changes about.
The only thing I haven't done yet is the suggestion to add more pictures.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
ngoldbaum: Sorry, that was a poor way to phrase it. You could add a comment here saying that doing `npy_intp n = dimensions[0]` means that this is a ufunc loop; a gufunc would process more than one element at a time.

ngoldbaum (approved):
Looking good now! It'd still be nice to have more visual aids but it's fine to not have any.

ikrommyd:
Thanks a lot for the review Nathan. if I like any visuals I create within the weekend I'll put them up to support the text. So please folks don't merge it yet :)

ikrommyd:
Apologies for the delay, moving took more effort than I expected it to 🤣. I added some svgs in a "handwritten" like style that I liked. Added minimal text explaining them (one sentence each).

ngoldbaum (commented):
The images look great and I think they'll help visual learners consume the post more easily.

ikrommyd:
Thanks Nathan. This should be ready from my side. If you or others want to review again, please go ahead :)

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,669 @@
...
+    in2_ptr += in2_step;
+    out_ptr += out_step;
+}
+```
+
+Each iteration casts the two input pointers to `double *` and dereferences them to get the current pair of values, adds them, writes the result through the output pointer, and then advances all three pointers by their strides to land on the next element.
+
+Here is the whole loop in one picture, with a strided view as the second input so that the strides are visible:
+
+![The three operands of the loop drawn as rows of cells, with a pointer at the current element of each row and an arrow for the stride to the next one. The second input is a strided view, so its stride skips a cell.](/posts/teaching-numpys-ufuncs-new-tricks/ufunc-loop-anatomy.svg)
```
melissawm:
```suggestion
![The three operands of the loop drawn as rows of cells, with a pointer at the current element of each row and an arrow for the stride to the next one. The first row shows a 4-element input array with a stride of one, showing an arrow between two consecutive elements. The second input is a strided view, so its stride skips a cell. The third row is the output and shows the result computed from the elements of both arrays recovered according to their strides.](/posts/teaching-numpys-ufuncs-new-tricks/ufunc-loop-anatomy.svg)
```

melissawm (approved):
@ikrommyd I have one quick suggestions for an alt-text that is entirely optional, otherwise LGTM. Thank you!! Ping me when you're done and I'll merge. Cheers!

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
ikrommyd: done, only changed "stride of one" - > "stride of one element" from your suggestion cause the former reads like one a one byte stride which would be false.

ikrommyd:
@melissawm - good to go from my side :) The published date however is still August 26 from when I started the PR, I don't know if you just wanna push a date change to my branch right before merging.

apps/labs/posts/teaching-numpys-ufuncs-new-tricks.md
```diff
@@ -0,0 +1,669 @@
+---
+title: "Teaching NumPy's ufuncs new tricks"
+authors: [iason-krommydas]
+published: August 26, 2026
```
melissawm:
```suggestion
published: September 6, 2026
```

melissawm:
Done - thank you @ikrommyd and reviewers!

PR #1000, opened 2026-09-03
Add "Finch meets MLIR" internship post

yusheng036:
## Text styling

- [x] The blog is written with [plain language](https://www.plainlanguage.gov/guidelines/) (where relevant).
- [x] If there are headers, they use the proper header tags in order to do so (with only one level-one header).
- [x] All links describe where they link to (for example, check the [Quansight labs website](https://labs.quansight.org/)).
- [x] Any kind of styling that the author uses (for example, bold for emphasis) is consistent throughout the blog.

## Non-text contents

- [x] Blog post featured image is in PNG or JPEG format, **not** SVG.
- [x] All content is represented as text (for example, images need alt text and videos need captions or descriptive transcripts).
- [x] If there are emojis, there are not more than three in a row.
- [x] Don't use [flashing gifs or videos](https://www.w3.org/TR/UNDERSTANDING-WCAG20/seizure-does-not-violate.html).
- [x] If it were to be read as plain text, the blog still makes sense and no information is missing.

PR #1001, opened 2026-09-04
Add MLIR code highlighting

mtsokol:
I would like to have full MLIR syntax support for the https://github.com/Quansight/Quansight-website/pull/1001 blogpost.


melissawm:
Hi @mtsokol - maybe we should stack this on #1000 to verify how it displays?

mtsokol:
If it can be in one batch I can combine it!

mtsokol:
Closing as it now lives in #1000

PR #1002, opened 2026-09-05
Add blog post "conda-build Adds Support for v1 Recipes"

jsmolic:
## Text styling

## Non-text contents


rgommers (commented):
@jsmolic this looks like a solid start. It does feel like it's quite short; the main thing that jumps out to me as missing is anything you did, or any particular challenges or things of interest in the implementation. We tend to write blog posts here that reflect work we did - and you did it, it's just not possible to see that from the to-the-point style right now.

PR #999, opened 2026-09-03
Quansight.com migration

mattip:
Migrate the 58 blog posts and 22 pages from quansight.com's wordpress site to this atro/next.js based site.

This is the first three commits: blog posts, pages and some general utilities, vercel preview extension.

Someone with vercel dashboard permissions needs to
- move the labs preview to the apps/labs/vercel.json lab-specific configuration
- add an additional preview pointing to apps/consulting/vercel.json

I added ignoreCommand directives to each preview file so changes in one app will not generate a preview in the other. That is why we need to move to the app-specific config for labs.

There is more to do. The page rendering is not great right now and here is a list of missing things:

- [x] 22 unimplemented widget instances still rendering as visible TODO placeholders:
     - form — 9 instances, 9 pages (needs a form backend decision: the live site posts to WP)
     - ~call-to-action — 6 instances, 4 pages~ done
     - price-table — 6 instances, 2 pages (training)
     - ~html — 1 instance~ done

- [x] 12 of 24 people use the grey Quansight logo as their photo instead of a real headshot/bio.
- [ ] 6 of 8 social icons are placeholders — 277-byte circle+initial SVGs (bluesky, facebook, instagram, linkedin, medium, youtube). Only github.svg and x.svg are real brand marks.
- [ ] 2 pages have no hero image, gradient-only fallback: open-source-integration (its per-page Elementor CSS 404s live) and subscribe-to-the-quansight-quarterly.

- [ ] No social tags at all on consulting — no og:/twitter: meta, no site in astro.config.mjs. See #994 
- [ ] No 404 page, no robots.txt, no sitemap.

PR #1004, opened 2026-09-09
Remove Labs vercel.json

trallard:
This is a prerequisite for #999.

Currently, we have a `vercel.json` file at the top of the repository. This is used as the default project configuration, overriding any configuration set at the app level or in the Vercel dashboard (which is currently causing the "consulting" deployment to fail).

For now, I have removed the configuration file and kept the deployment configuration in the Vercel dashboard. If we need more complex configurations than the deployment build, we can add a new `vercel.json` in the `apps/labs` directory.

### DIFF
--- a/apps/labs/pages/home.yml
+++ b/apps/labs/pages/home.yml
@@ -18,7 +18,7 @@ blocks:
     and open source projects and communities they rely on.
   rightColumn: We are a team of developers, community leads, designers, and writers. Together we create,
     contribute to, and maintain open source technology for research and data science workflows. We help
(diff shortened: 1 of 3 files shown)


trallard:
@pavithraes this is now working for Labs. Can you please review and merge if all looks ok?

Note that the consulting build will fail until #999 is merged. I can remove the project for now if y'all think is too noisy.

pavithraes:
Thanks for this!

>  can remove the project for now if y'all think is too noisy

I think it's OK to keep it, especially since we are planning to get #999 done asap

pavithraes (approved):
LGTM, thank you!

PR #1006, opened 2026-09-12
Add "How to fit a Smooth Curve? Using SciPy" Internship 2026 blog 

aadya940:
## Text styling

## Non-text contents


rgommers:
I had a quick read through, looks like a nice story @aadya940!

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+</div>
+
+The mathematics of this was in fact later worked out in [this paper](https://pub.math.leidenuniv.nl/~smitbde/papers/2003-de_smit-lenstra-escher.pdf).
+So hopefully, I convinced you smooth curves are interesting.
+
+## But, how do you fit a smooth curve?
+
+Notice, the term "fit". Say you made a bunch of measurements, at each input <i>xi</i> you recorded a value <i>yi</i>, and the recordings carry some
+noise. We want a smooth curve that follows these data points <i>(xi, yi)</i> without chasing the noise. We can fit one by minimizing this objective function:
```
melissawm: Why not use math here?
```suggestion
Notice the term "fit". Say you made a bunch of measurements, at each input $x_i$ you recorded a value $y_i$, and the recordings carry some
noise. We want a smooth curve that follows these data points $(x_i, y_i)$ without chasing the noise. We can fit one by minimizing this objective function:
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+<div align="center">
+
+
+<img width="792" alt="The objective J equals the sum of squared residuals plus lambda times the integral of the squared second derivative" src="/posts/fit-a-smooth-curve-using-scipy/objective-function.png" />
+
```
melissawm: You could also replace the image with math directly here.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+When you optimise $J$ over the space of all smooth functions, you get a Spline. For a more detailed derivation, consult the companion write-up 
+[I wrote during the internship](https://github.com/aadya940/scipy-bspline-testing/blob/main/B_Splines_with_arbitary_knots-gcv.pdf).
+
+
+## Splines
+
+Splines are piecewise cubic polynomials. The domain of each piece is bounded by a pair of real numbers, let's call the sorted list of these
+real numbers $t$. These are called the knots, and the simplest way to think about them is, knots are the places where you allow the curve to bend.
+
+It is a special type of piecewise polynomial function made by joining individual cubic (third-degree) polynomials together smoothly between a set of data points or knots. Instead of fitting a single, high-degree polynomial across an entire data set, which can cause wild oscillations known as Runge's phenomenon, splines use lower-degree cubic polynomials on smaller subintervals while maintaining continuous transitions from one piece to the next.
```
melissawm: The subject is not super clear here. I'd go with

```suggestion
Splines are a special type of piecewise polynomial function made by joining individual cubic (third-degree) polynomials together smoothly between a set of data points or knots. Instead of fitting a single, high-degree polynomial across an entire data set, which can cause wild oscillations known as [Runge's phenomenon](https://en.wikipedia.org/wiki/Runge%27s_phenomenon), splines use lower-degree cubic polynomials on smaller subintervals while maintaining continuous transitions from one piece to the next.
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+
+<img width="900" alt="Left: a single degree-12 polynomial through 13 points oscillates wildly near the edges. Right: a cubic spline through the same points stays calm, drawn in a pencil sketch style." src="/posts/fit-a-smooth-curve-using-scipy/piecewise-vs-polynomial.png" />
+
+
+</div>
+
+<p align="center"><i><small>Same 13 points in both panels.</small></i></p>
+
+So you can fit a smooth curve through a bunch of points using a spline and use $\lambda$ to control the amount of smoothing. With $\lambda = 0$, the curve may freely interpolate the points, with $\lambda = \infty$ you get a straight line. Using this concept, you can start your artwork. Note that an image is just a bunch of horizontal or vertical lines stacked in order. So you can fit splines with separate $\lambda$'s and create images with focus on one particular object. Here is an example,
```
melissawm:
```suggestion
So you can fit a smooth curve through a bunch of points using a spline and use $\lambda$ to control the amount of smoothing. With $\lambda = 0$, the curve may freely interpolate the points, with $\lambda = \infty$ you get a straight line. Using this concept, you can start your artwork. Note that an image is just a bunch of horizontal or vertical lines stacked in order. So you can fit splines with separate $\lambda$'s and create images with focus on one particular object. Here is an example:
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+<div align="center">
+
+
+<img width="700" alt="Artwork of a building and sky drawn entirely with smoothing splines fitted line by line" src="/posts/fit-a-smooth-curve-using-scipy/spline-art-1.png" />
```
melissawm, line 112: I am not that young anymore, so the caption on the images are reeeeeally small 😅 any chance they can be increased?

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+
+</div>
+
+Or you can use Splines, as a paintbrush, here's another example:
```
melissawm: Can you explain a bit more? I know we don't want to make this super detailed but the paint strokes in the image are segments and I'm not sure I understand what you meant by "splines as a paintbrush". In any case, if you keep this sentence, I'd suggest


```suggestion
Or you can use Splines as a paintbrush. Here's another example:
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+<div align="center">
+
+
+<img width="700" alt="Artwork drawn using splines as paintbrush strokes" src="/posts/fit-a-smooth-curve-using-scipy/spline-art-2.png" />
+
+
+</div>
+
+These have been built using <b>scipy.interpolate.make_smoothing_spline</b> function provided by the SciPy project which I had the pleasure to work on as an Intern (Summer, 2026) at Quansight Labs under [Evgeni Burovski](https://github.com/ev-br) and [Gagandeep Singh](https://github.com/czgdp1807). Huge Shoutout to them!
```
melissawm:
```suggestion
These have been built using [scipy.interpolate.make_smoothing_spline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.make_smoothing_spline.html) function provided by the SciPy project which I had the pleasure to work on as an Intern (Summer, 2026) at Quansight Labs under [Evgeni Burovski](https://github.com/ev-br) and [Gagandeep Singh](https://github.com/czgdp1807). Huge shoutout to them!
```

melissawm: Best practice is to add a link or citation at the first mention of that concept, so I'd move the link from the next section to here.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+[I wrote during the internship](https://github.com/aadya940/scipy-bspline-testing/blob/main/B_Splines_with_arbitary_knots-gcv.pdf).
+
+
+## Splines
+
+Splines are piecewise cubic polynomials. The domain of each piece is bounded by a pair of real numbers, let's call the sorted list of these
+real numbers $t$. These are called the knots, and the simplest way to think about them is, knots are the places where you allow the curve to bend.
+
+It is a special type of piecewise polynomial function made by joining individual cubic (third-degree) polynomials together smoothly between a set of data points or knots. Instead of fitting a single, high-degree polynomial across an entire data set, which can cause wild oscillations known as Runge's phenomenon, splines use lower-degree cubic polynomials on smaller subintervals while maintaining continuous transitions from one piece to the next.
+We're only assuming the cases that work on a <i>two-dimensional</i> plane for now.
```
melissawm, line 93: I think you don't need to mention the knots just yet - it's not necessary for the explanation you are working through, and creates a little bit of confusion. You also repeat the same idea a few times in these two paragraphs, so here's a more concise suggestion:

```suggestion
Splines are a special type of piecewise polynomials. Instead of fitting a single, high-degree polynomial across an entire data set, which can cause wild oscillations known as [Runge's phenomenon](https://en.wikipedia.org/wiki/Runge%27s_phenomenon), splines use lower-degree cubic polynomials on smaller subintervals while maintaining continuous transitions from one piece to the next.
We're only assuming the cases that work on a <i>two-dimensional</i> plane for now.
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+It was internally assumed that $t$, the knots, are always equal to $x$, the data sites, so it was not in the API. This is a popular choice since there is a classical theorem that when you set $t$ = $x$ you always get the curve that best minimizes the objective function we discussed earlier.
+
+But there is a tradeoff hiding in that choice. Internally, SciPy builds matrices to solve this, and the size of the system depends on the number of pieces in the piecewise function. With $t$ = $x$ the number of pieces grows with the data. For `1,000,000` datapoints, the size of the matrix is `1,000,000 x 1,000,000`. Hence, some people may want to pass a smaller $t$ to `make_smoothing_spline` to get a slightly worse curve but much faster. This was not available in SciPy yet, I worked through the summer to add it.
+
```
melissawm: For your consideration: here's a different way to introduce the knots. Feel free to take it or leave it. My goal is basically to chain the ideas in a more direct way - building over the simpler choice and explaining the problem before mentioning the alternative of knots.

```suggestion
The SciPy library provides a number of routines to work with different kinds of splines. The spline we discussed is provided by [scipy.interpolate.make_smoothing_spline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.make_smoothing_spline.html). It takes in the $x$, $y$ and $\lambda$ values and returns $f$, the fitted spline.

In the simplest case, we assume that the spline will interpolate $x$, the data sites. But there is a tradeoff hiding in that choice. Internally, SciPy builds matrices to solve this, and the size of the system depends on the number of pieces in the piecewise function. For `1,000,000` datapoints, the size of the matrix is `1,000,000 x 1,000,000`. Hence, some people may want to pass a smaller number of points, or *knots* (denoted by $t$) to `make_smoothing_spline` to get a slightly worse curve but much faster. This was not available in SciPy yet, I worked through the summer to add it.

```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+On the good side, these matrices are often banded (<i>Because Piecewise, Only a few diagonals are non-zero </i>). The design matrix $X$ (defined below) has at most `k + 1` non-zeros per row, where `k` is the degree of each piece, and the matrix we solve, $X^\top W X + \lambda\,\Omega$ (also defined below), is banded too, with `2k + 1` diagonals. Here, since $k$ is just $3$, the number of non-zero elements is $(k + 1) * (num. datapoints)$, in this case it's $4,000,000$.
```
melissawm:
```suggestion
On the good side, these matrices are often banded (<i>only a few diagonals are non-zero </i>). The design matrix $X$ (defined below) has at most `k + 1` non-zeros per row, where `k` is the degree of each piece, and the matrix we solve, $X^\top W X + \lambda\,\Omega$ (also defined below), is banded too, with `2k + 1` diagonals. Here, since $k$ is just $3$, the number of non-zero elements is $(k + 1) * (num. datapoints)$, in this case it's $4,000,000$.
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+By the way, the matrix I'm talking about is called the "design matrix", It is built from the values of the basis elements at the data points. We solve the equation $y = X * c$ in the least squares sense (that is the $\lambda = 0$ limit of our objective), where $c$ are some coefficients that scale or squish the influence of some columns of the design matrix. And since each basis element is non-zero over only a few pieces, the fewer the knots, the smaller the dimension of $X$. Expanding it further, the closed-form solution to our minimization problem is given by this matrix equation:
```
melissawm:
```suggestion
By the way, the matrix I'm talking about is called the "design matrix". It is built from the values of the basis elements at the data points. We solve the equation $y = X * c$ in the least squares sense (that is the $\lambda = 0$ limit of our objective), where $c$ are some coefficients that scale or squish the influence of some columns of the design matrix. And, since each basis element is non-zero over only a few pieces, the fewer the knots, the smaller the dimension of $X$. Expanding it further, the closed-form solution to our minimization problem is given by this matrix equation:
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,361 @@
...
+
+$$
+(X^\top W X + \lambda\,\Omega)\,\boldsymbol{c} \;=\; X^\top W \boldsymbol{y},
+$$
+
+Here, $W$ is a diagonal matrix of optional per-point weights, the identity matrix if you don't pass any, and the new $\Omega$ is just the matrix form of the integral of the squared second derivative we saw earlier. <br /> Now we have two options, either differentiate and integrate numerically every time to compute $\Omega$ or derive a general matrix form of $\Omega$ which circumvents this procedure, and once you have matrices with nice properties you can apply optimizations on them, inspect them etc. <br /> So what is the matrix form and how to compute it? <br /> That was about half of my internship. Deriving $\Omega$ using papers and books going back to the 1980's. Other implementations like the R programming language's libraries are GPL licensed, so we deliberately did not look at their source code (only used their numerical output as a black-box check). Apart from that, most of MATLAB, Octave, Julia etc. don't support user defined knot vectors either.
```
melissawm: These `<br>` here seem out of place to me - the formatting in the blog post doesn't look as good. How about making them into real paragraph breaks or removing them entirely?

melissawm (commented):
Hi @aadya940 ! I gave a first pass looking mainly at the technical writing aspect of it. I left a few suggestions for reorganizing some of the content for clarity and also some stylistic/grammar notes. Let me know if anything is not clear!

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: Thanks! I will not use this one because there is a slight misunderstanding. At `lam=0`, the spline may interpolate if it has enough degrees of freedom, otherwise it will not, and for `lam > 0` it smooths rather than interpolates. Also, this makes it sound like the knots need to be a subset of, or equal to, the data points, when in reality they are completely independent of the data points and can be totally different. But your point regarding the flow was well taken, so I will change it along those lines.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: I added a small paragraph, I agree this does needs an explanation!

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: Thanks! This indeed does look better. How did I miss it? Huh 😮‍💨

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: done!

aadya940:
Thanks @melissawm for the quick and detailed review! I'm really famous for adding unnecessary commas, but I did fix that 😆. Looks much better now indeed.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+<i>Here is the spline it generates: </i>
+
+
+<img width="880" alt="A free least squares spline and a clamped one whose endpoints are pinned to zero, shown with stars" src="/posts/fit-a-smooth-curve-using-scipy/clamp-values.png" />
+
+
+The main implementation task was to solve a smaller system. A spline fitting problem has $m$ basis functions, where $m$ = len($t$) $- 4$, and each 
+basis function gets one coefficient, so we solve for a coefficient vector $c$ of length $m$. Now, at a clamped boundary only the first basis function is nonzero, 
+with $B_0(x[0]) = 1$, so the spline evaluated at the first point is simply $c_0$. The constraint spl($x[0]$) $= c$<i>i</i> therefore fixes $c_0 = c$<i>i</i> directly, and likewise 
+$c_{m-1} = c$<i>f</i> at the other end. So the clamp values feature effectively has to solve for only $m - 2$ coefficients, the two boundary ones are known before the solve even starts.
```
czgdp1807: **Question** - The `c$<i>i</i>` and `c$<i>f</i>` appear as `ci` and `cf` respectively. Was this your intention? Or did you want to subscript `i` and `f` as follows,


```suggestion
with $B_0(x[0]) = 1$, so the spline evaluated at the first point is simply $c_0$. The constraint spl($x[0]$) $= c_i therefore fixes $c_0 = c_i directly, and likewise 
$c_{m-1} = c_f at the other end. So the clamp values feature effectively has to solve for only $m - 2$ coefficients, the two boundary ones are known before the solve even starts.
```

apps/labs/public/posts/fit-a-smooth-curve-using-scipy/hero.png
czgdp1807, line 1: This image appears to be exactly same as - `apps/labs/public/posts/fit-a-smooth-curve-using-scipy/featured.png`. Do we still need it? If yes, then what is the reason? If no then let's delete one of those.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
+---
+title: 'How to fit a Smooth Curve? Using SciPy'
```
czgdp1807, line 2: The question mark in between is bothering me a little bit. As far as I checked it might be grammatically incorrect as well. Here are two suggestions from my end,

- How to fit a smooth curve using SciPy
- Fitting smooth curves with SciPy

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+These are smooth because their derivatives are continuous. The smoother the curve, the higher the order up to which its derivatives stay continuous.
+For example, we know intuitively that this is smooth:
+
+$$
+y = \sin(x)
+$$
+
+But mathematically you can differentiate $y = \sin(x)$ an infinite number of times and still get a continuous curve. More importantly, if the
+derivatives were discontinuous, the path would have sharp edges and applications like the roller coaster ride wouldn't be suitable. But smooth is underrated, mostly we're always given examples of motion for mathematical smoothness but it plays a huge role in aesthetics. There was a popular
```
czgdp1807, line 54: Does it makes sense to rewrite, "But smooth is underrated, mostly we're always given examples of motion for mathematical smoothness but it plays a huge role in aesthetics" into something like,

```suggestion
derivatives were discontinuous, the path would have sharp edges and applications like the roller coaster ride wouldn't be suitable. But smoothness isn't only about motion, it matters a lot for aesthetics too. There was a popular
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+I spent some parts of the Internship on Optimizations. A few examples:
+- Whenever a matrix is a special type of matrix called a [Hermitian matrix](https://en.wikipedia.org/wiki/Hermitian_matrix) (symmetric, in our real-valued case) and positive definite, you can use SciPy's `solveh_banded` solver which is several times faster than the general `solve_banded` solver.
+- Whenever you have a banded matrix (<i> which is nearly always in this case </i>), use SciPy's [diags_array](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.diags_array.html) for optimized routines.
```
czgdp1807, line 162: **Optional** - Better to follow one convention of spelling words, though not strict at all.

```suggestion
- Whenever you have a banded matrix (<i> which is nearly always in this case </i>), use SciPy's [diags_array](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.diags_array.html) for optimised routines.
```

Taking the convention from your above sentence,

> When you optimise $J$ over the space of all smooth functions, you get a Spline. For a more detailed derivation, consult the companion write-up

czgdp1807 (commented):
This is my first round of review. Will do another round for checking the semantics. TYSM for the PR and your overall work. 🤝 ❤️

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+- <b> You decide where the curve bends. </b> Since knots are the bending points, you can place more of them in the regions where the signal has features and fewer in the flat regions. All the artwork in this post was made this way.
+- <b> It is faster even in the old case. </b> The new path solves its system with `solveh_banded`, which exploits symmetry, instead of the more general `solve_banded`. In our benchmarks this alone made it about 2x faster than the old `t = x` path on the same problem.
+-  $\lambda$ <b>= None still works. </b> The GCV search was re-derived for arbitrary knots, so automatic smoothing is not a $t = x$ exclusive anymore.
+- <b> Nothing changes if you don't opt in. </b> Leave $t$ out and you get exactly the old behavior. In fact the test suite demands it: with knots at the data and the same $\lambda$, the new path reproduces the old one to about `1e-12`, and the penalty matrix $\Omega$ is computed in closed form, no quadrature error to budget for.
```
czgdp1807: > In fact the test suite demands it: with knots at the data and the same $\lambda$ the new path reproduces the old one to about `1e-12`

- The above line suggests that the test you are talking about is, `xp_assert_close(spl1.c, spl2.c, atol=1e-12 * max(1.0, lam / 1e-4))`. 
- The claim is, "the new path reproduces the old one to about `1e-12`" given new knots and the `lambda` are same as those used in older path.
- However, `atol=1e-12 * max(1.0, lam / 1e-4)` suggests that the claim may not always be correct. The `atol` factor scales with `lam` (the higher the `lam` value, the more the new path deviates from the older one). See below,

```python
>>> lam = [1e-4, 0.5, 100.0]
>>> atols = [max(1.0, l/1e-4) * 1e-12 for l in lam] 
>>> atols
[1e-12, 5e-09, 1e-06]
```

What do you say? If you find any incorrectness in any of my points above, please point it in a reply to this review comment.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+spl = make_smoothing_spline(x, y, lam=lam, t=my_knots)
+
+# or let GCV pick lambda on your knots, that works too
+spl = make_smoothing_spline(x, y, t=my_knots)
+```
+
+What this buys you:
+- <b> The solve is sized by your knots, not your data. </b> Remember the `1,000,000 x 1,000,000` matrix from before? Pass 100 knots and the system is roughly `100 x 100` (banded, on top of that), no matter how many points you have. The data only enters through small products like $X^T y$.
+- <b> You decide where the curve bends. </b> Since knots are the bending points, you can place more of them in the regions where the signal has features and fewer in the flat regions. All the artwork in this post was made this way.
+- <b> It is faster even in the old case. </b> The new path solves its system with `solveh_banded`, which exploits symmetry, instead of the more general `solve_banded`. In our benchmarks this alone made it about 2x faster than the old `t = x` path on the same problem.
```
czgdp1807: > In our benchmarks

Can you link the benchmarks here?

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+I spent some parts of the Internship on Optimizations. A few examples:
+- Whenever a matrix is a special type of matrix called a [Hermitian matrix](https://en.wikipedia.org/wiki/Hermitian_matrix) (symmetric, in our real-valued case) and positive definite, you can use SciPy's `solveh_banded` solver which is several times faster than the general `solve_banded` solver.
```
czgdp1807: > SciPy's `solveh_banded` solver which is several times faster than the general `solve_banded` solver.

"several times faster" sounds a little vague to me. Also I have some questions about this claim,

- Is `solveh_banded` always faster than `solve_banded`?
- If yes, then where can I find support for this claim?
- If no, then is `solveh_banded` faster than `solve_banded` for our use case? If yes, then why? A single reason would satiate me.

czgdp1807 (commented):
Leaving some questions/comments. TY.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: Thanks, great catch, yes this is a rendering bug I will fix it.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: It's intentional. Grammatically, it is pretty common to have such titles. A small question followed by an answer and the blog post answers why. I'd keep this if it doesn't sound like a major blocker to you.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: Thanks for the suggestion, yes, "optimized" is more widely used so I'd keep it consistent throughout.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: Yes but this just needs a wording rewrite, for example:
```
to about 1e-12 at moderate λ (the tolerance scales with λ, as conditioning does)
```

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: `solveh_banded` is for symmetric positive definite matrices, and in that case it is faster essentially by construction, Cholesky does about half the flops of LU exploiting symmetricity. That's exactly our case, `X.T W X + lam * Omega` is symmetric positive definite for `lam > 0`, which is why we use it. The reference for the claim is the factorization cost itself rather than a benchmark. So, several times faster can be changed to `>= 2x faster`.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
czgdp1807: > the tolerance scales with λ, as conditioning does

This would indeed make things transparent.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: They were local runs during development and I don't have them published in a linkable form. I could remove this claim if it sounds unfair or write a link to the benchmarks somewhere or maybe just keep as it is for it may still be obvious that `solveh_banded` is 2x faster than `solve_banded` due to symmetricity as answered below, What do you think?

apps/labs/public/posts/fit-a-smooth-curve-using-scipy/hero.png
aadya940: Thanks, Both of them are needed, `featuredImage` is used for the blog card and social previews and hero is the banner above the post. What I try to do is point both of them to the same image.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
aadya940: At some places the tone is deliberately informal considering it's a blog post. However, I do not have strong opinions about this one and happy to change it. Just curious, what's the issue here?

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
czgdp1807: > write a link to the benchmarks somewhere

Personally I'd prefer this. Something like, `[In our benchmarks](...)`

---

> may still be obvious that `solveh_banded` is 2x faster than `solve_banded` due to symmetricity as answered below

By doing so, we are relying on the presumption that the reader will be familiar of the fact that `solveh_banded` is 2x faster than `solve_banded`. If we are fine with this presumption then keep it as is but remove the "In our benchmarks".

My opinion - By reading the line, "In our benchmarks this alone made it about 2x faster than the old `t = x` path on the same problem." - a reader would naturally want to see the exact benchmarking code. If you don't want to link it, then the other way is to derive "about 2x faster than the old `t = x` path" from the fact "`solveh_banded` is 2x faster than `solve_banded`" and provide evidence ([theoretical](https://share.google/aimode/8o9sXIL9x6f1C0Lr8) or practical - I leave it upto you) for the latter.

Linking benchmarks is easier I think.

aadya940:
Thanks @czgdp1807, this was really helpful! I’ll apply some of the changes in the next commit, and I’ve replied to the others in the review.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
czgdp1807: > Just curious, what's the issue here?

The sentence is a bit too long to read. You can make it shorter or break it into smaller ones (applies in general too but I wouldn't push too much on it. It's just a blog post).

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+Notice, the term "fit". Say you made a bunch of measurements, at each input $x_i$ you recorded a value $y_i$, and the recordings carry some
+noise. We want a smooth curve that follows these data points $(x_i, y_i)$ without chasing the noise. We can fit one by minimizing this objective function:
+
+$$
+J[f] = \sum_{i=1}^{n} \big( y_i - f(x_i) \big)^2 \; + \; \lambda \int_{x_1}^{x_n} \big( f''(u) \big)^2 \, du
+$$
+
+Here, $f(x)$ is the fitted curve, $f''$ is the second derivative of the same curve which measures how curvy or jumpy the curve
+is. Hence, the first term in $J$ is how closely the curve fits the data, the second term is how jumpy/curvy the curve is and
+$\lambda$ controls the amount of penalty to apply for bending.
```
ev-br, line 82: Consider discussing the limiting cases here: `\lambda \gg 1` is a straight line, `lambda = 0` is a unpenalized LSQ fit which interpolates the data points if `f(x)` is flexible enough. You can also drop the _ridge regression_ term here, so that readers make the connection.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+The direction changes slightly at every step, so the path we get is a little shaky. So we fit a smoothing spline on these points to get a smooth curve through them, and we draw this curve as one thick stroke in the color the photo has at that spot. We repeat this from 9000 random spots, and all the strokes together form the full picture.
+Nothing here blurs or filters the photo, we are just repainting it stroke by stroke, and each stroke is a fitted spline.
+
+These have been built using the [scipy.interpolate.make_smoothing_spline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.make_smoothing_spline.html) function provided by the SciPy project which I had the pleasure to work on as an Intern (Summer, 2026) at Quansight Labs under [Evgeni Burovski](https://github.com/ev-br) and [Gagandeep Singh](https://github.com/czgdp1807). Huge shoutout to them!
```
ev-br: What are "these" here---now that I read about pantbrushes and a pagoda drawing, it reads as if that's what you've been building during the internship.

Also, have these pagoda drawings actually been built with `make_smoothign_spline`? If so, do add a link to code which generated the drawings!
(otherwise, detail how these were drawn, where they were taken from)

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+## SciPy
+
+The SciPy library provides a number of routines to work with different kinds of splines. The spline we discussed is provided by `scipy.interpolate.make_smoothing_spline`, it takes in the $x$, $y$ and $\lambda$ values and returns $f$, the fitted spline. But wait, where does one piece of the piecewise cubic end and the next one begin? These joining positions are called the <i>knots</i>, we collect them in a sorted list and call it $t$. The simplest way to think about them is, knots are the places where you allow the curve to bend. So where does $t$ come from?
+
+It was internally assumed that the knots are always placed at $x$, the data sites, so $t$ was not in the API. This is a popular choice since there is a classical theorem that when you place the knots at the data sites you always get the curve that best minimizes the objective function we discussed earlier.
```
ev-br: While there's no need for the last bit of math rigor in a blog post, "best minimizes the objective function"---what is "best" here, in what space etc. Consider rephrasing, maybe just drop the second sentence, just state that a classic smoothing spine uses all data sites, and in the next paragraph just start from the design matrix sizes, and state that a smaller number of judiciously chosen knots gives essentially the same result, with much smaller computational cost.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+But there is a tradeoff hiding in that choice. Internally, SciPy builds matrices to solve this, and the size of the system depends on the number of pieces in the piecewise function. With $t$ = $x$ the number of pieces grows with the data. For `1,000,000` datapoints, the size of the matrix is `1,000,000 x 1,000,000`. Hence, some people may want to pass a smaller $t$ to `make_smoothing_spline` to get a slightly worse curve but much faster. This was not available in SciPy yet, I worked through the summer to add it.
+
+On the good side, these matrices are often banded (<i>only a few diagonals are non-zero </i>). The design matrix $X$ (defined below) has at most `k + 1` non-zeros per row, where `k` is the degree of each piece, and the matrix we solve, $X^\top W X + \lambda\,\Omega$ (also defined below), is banded too, with `2k + 1` diagonals. Here, since $k$ is just $3$, the number of non-zero elements is $(k + 1) * (num. datapoints)$, in this case it's $4,000,000$.
```
ev-br: Consider rephrasing this paragraph:
- what is "often" 
- why is k=3 where does that come from

I think you can just state that the most common choice is cubic splines (which are piecewise cubic polynomials, `k=3`), the design matrices are banded, and continue as you were.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+By the way, the matrix I'm talking about is called the "design matrix". It is built from the values of the basis elements at the data points. We solve the equation $y = X * c$ in the least squares sense (that is the $\lambda = 0$ limit of our objective), where $c$ are some coefficients that scale or squish the influence of some columns of the design matrix. And, since each basis element is non-zero over only a few pieces, the fewer the knots, the smaller the dimension of $X$. Expanding it further, the closed-form solution to our minimization problem is given by this matrix equation:
```
ev-br: I'm thoroughly confused now. There's a penalized regression, there's somethign called X, which solves some equation, but no, it solves it in a least squares sense, and then there's lambda=0, so are we only solving with lambda=0?

Basically, either the reader knows what's going on already, and then the discussion is not needed; or the reader does not know it, and then it's confusing.

Consider introducing  `f(x) = \sum_j c_j B_j(x)` where `B_j(x)` are the so-called b-splines, and then you have the design matrix `X_{ij} = B_j(x_i)`, and then the matrix form of the solution follows. A reader with the knowledge glances across it; a reader who does not gets the main idea, has the terminology to look up the details.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+$$
+(X^\top W X + \lambda\,\Omega)\,\boldsymbol{c} \;=\; X^\top W \boldsymbol{y},
+$$
+
+Here, $W$ is a diagonal matrix of optional per-point weights, the identity matrix if you don't pass any, and the new $\Omega$ is just the matrix form of the integral of the squared second derivative we saw earlier. 
+
+Now we have two options, either differentiate and integrate numerically every time to compute $\Omega$ or derive a general matrix form of $\Omega$ which circumvents this procedure, and once you have matrices with nice properties you can apply optimizations on them, inspect them etc.
```
ev-br: To construct the `Omega` matrix, we have two options....

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+So what is the matrix form and how to compute it? 
+
+That was about half of my internship. Deriving $\Omega$ using papers and books going back to the 1980's. Other implementations like the R programming language's libraries are GPL licensed, so we deliberately did not look at their source code (only used their numerical output as a black-box check).
```
ev-br: 1. So which of the two options from the previous paragraph was it?

2. Consider discussing other implementations separately (with a nice table you used elsewhere). Licensing concerns belong to taht discussion, too. Here what is needed is that you derived the explicit form of Omega, and validated it against R, I think.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+SciPy did use matrices but they assume the $t$ = $x$ thing and hence, their matrices are more specialized than what the general case needs. The complete derivation is available in [my companion write-up](https://github.com/aadya940/scipy-bspline-testing/blob/main/B_Splines_with_arbitary_knots-gcv.pdf). Also, when $t$ = $x$ the design matrix was a square matrix, now it is not, so we can't apply some shortcuts we applied earlier. Furthermore, the fitted curve depends on the value of $\lambda$, so what is a good value of $\lambda$? The classic procedure is leave-one-out cross validation, drop one point, fit on the rest, check how well the fit predicts the dropped point, and repeat this for every point and every candidate $\lambda$. Accurate, but costly, you refit $n$ times per candidate. Generalized Cross Validation (GCV) is a shortcut that estimates the same score from a single fit. So previously in SciPy you could pass $\lambda$ = None and SciPy would automatically find the appropriate one using GCV. This algorithm assumed $t$ = $x$ too, so I went on to derive the general case for that as well.
```
ev-br: 1. Will need to change the link to the stable doi or arxiv, when available.

EDIT: these two point below have been addressed.

2. Consider making the GCV discussion more prominent. 

3. Omega / X being square or rectangular : you worked through a generalization, can brag about it a bit

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+I spent some parts of the Internship on Optimizations. A few examples:
```
ev-br: I typed a long comment here only to realize that Optimizations here meant to mean code optimizations not mathematical optimization (and since I just read about minimizing a cost function, teh reader in me expected the latter).

Consider rephrasing this paragraph. One way is to say that the matrices have special structure (banded, symmetric), and your implementation used the structure to great effect.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+# or let GCV pick lambda on your knots, that works too
+spl = make_smoothing_spline(x, y, t=my_knots)
+```
+
+What this buys you:
+- <b> The solve is sized by your knots, not your data. </b> Remember the `1,000,000 x 1,000,000` matrix from before? Pass 100 knots and the system is roughly `100 x 100` (banded, on top of that), no matter how many points you have. The data only enters through small products like $X^T y$.
+- <b> You decide where the curve bends. </b> Since knots are the bending points, you can place more of them in the regions where the signal has features and fewer in the flat regions. All the artwork in this post was made this way.
+- <b> It is faster even in the old case. </b> The new path solves its system with `solveh_banded`, which exploits symmetry, instead of the more general `solve_banded`. Cholesky does about half the flops of LU, so this alone makes it more than or equal to 2x faster than the old `t = x` path on the same problem.
+-  $\lambda$ <b>= None still works. </b> The GCV search was re-derived for arbitrary knots, so automatic smoothing is not a $t = x$ exclusive anymore.
```
ev-br: lambda=None is implementation details of how to use GCV. And BTW, did you ever write above that you worked through the derivation and made sure that the GCV computation works for user-supplied knots?

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+- <b> Nothing changes if you don't opt in. </b> Leave $t$ out and you get exactly the old behavior. In fact the test suite demands it: with knots at the data and the same $\lambda$, the new path reproduces the old one to about `1e-12` at moderate $\lambda$ (the tolerance scales with $\lambda$, as conditioning does), and the penalty matrix $\Omega$ is computed in closed form, no quadrature error to budget for.
+
+The feature lives in [scipy/scipy#25862](https://github.com/scipy/scipy/pull/25862) and the GCV follow-up is in review in [scipy/scipy#26141](https://github.com/scipy/scipy/pull/26141), so it makes its way to a SciPy release near you soon.
+
+### Tutorial 1: Lots of points, few knots
```
ev-br: Nit: demo 1?

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+</div>
+
+The data here is generated as function + noise, and the ground truth is just the function, without the noise. The RMSE is the root mean squared error between the fitted curve $f$ and 
+the ground truth $g$,
+
+$$
+\text{RMSE} = \sqrt{\frac{1}{n} \sum_i \big(f(x_i) - g(x_i)\big)^2}
+$$
+
+so intuitively, it is the typical distance between the curve we fit and the curve we were trying to recover. The reference value to compare against is the noise level, the noise added to the data has standard deviation `0.05`. Same data, same $\lambda$, same number of coefficients, same everything, only the placement differs. The uniform knots oversmooth the bump and wiggle around in the quiet region, giving RMSE `0.053`, the fit misses the truth by about as much as the noise itself, so it recovered nothing beyond the data. The placed knots are clustered around the bump, so the bump comes out clean and the rest stays calm, giving RMSE `0.013`, four times below the noise level, so the fit genuinely recovers the underlying function. The small tick marks in the plots show where the knots are.
```
ev-br: Consider placing 

> The small tick marks in the plots show where the knots are

at the start of the paragraph.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+By the way, if you don't want to hand-pick the knots, NumPy gives you an easier option:
+
+```python
+>>> t = clamp(np.quantile(x, np.linspace(0, 1, 10)[1:-1], method="nearest"))
+```
+
+This places the knots at the quantiles of your data, so they end up dense wherever your data points are dense and sparse where they are sparse. The `method="nearest"` part makes every knot
+exactly on one of your data sites, which guarantees each knot interval actually contains data, so the fit is always well posed.
```
ev-br: So the take-home message for me as a reader is that placing knots away from data makes the fit ill-posed?

Consider rephrasing:

1. If you don't want to hand-pick the knots, a simple choice is to place them at quantiles. 
2. If you additionally want the knots to be at the data sites, NumPy gives you an easy option of `method="nearest"`

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+</div>
+
+On the left is what happens internally. GCV asks, for each candidate $\lambda$,
+how well would the fit predict each point if that point were left out, and picks
+the minimum. Note the x-axis, it's $\log_{10}(\lambda / r)$, not $\log_{10}(\lambda)$.
+That ratio $r$ is what makes the search work in any unit system. $\lambda$ has units,
+so if you rescale $x$, the right $\lambda$ changes too. You can check this: I fit
+the same measurements with $x$ in metres and in kilometres, and GCV chose
+`2.40e-03` and `2.40e-12`, a factor of $10^9$ apart, exactly the $(10^3)^3$ the
```
ev-br: nit: `$2.4\times 10^{-3}$`, also on the plots

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+<img width="1590" alt="The GCV criterion versus the dimensionless smoothing parameter with a clear minimum, and the resulting fit" src="/posts/fit-a-smooth-curve-using-scipy/tutorial-3-gcv.png" />
+
+
+</div>
+
+On the left is what happens internally. GCV asks, for each candidate $\lambda$,
+how well would the fit predict each point if that point were left out, and picks
+the minimum. Note the x-axis, it's $\log_{10}(\lambda / r)$, not $\log_{10}(\lambda)$.
+That ratio $r$ is what makes the search work in any unit system. $\lambda$ has units,
```
ev-br: The reader is now wondering where does `r` come from.

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+the same measurements with $x$ in metres and in kilometres, and GCV chose
+`2.40e-03` and `2.40e-12`, a factor of $10^9$ apart, exactly the $(10^3)^3$ the
+units demand, and the two fitted curves are the same curve. The search is over
+the dimensionless variable, so the fixed window works for everyone. This also
+addresses a long-standing SciPy issue where the old search failed on data
+with large or small $x$ scales.
+
+## Beyond the main feature
+
+The knots work was the centerpiece, but a few other things happened around it:
```
ev-br: the knots work was the centerpiece _of the internship_?

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+Here is the PR: [scipy/scipy#26168](https://github.com/scipy/scipy/pull/26168)
+
+## Future Work
+
+SciPy still needs work on some features users may want, and I'm planning to work on them after the Quansight Internship. Here are a few likely candidates:
+
+- P-spline penalties, a cheaper discrete alternative to the exact integral penalty
+- Higher order derivative penalties, penalize the third derivative instead of the second when you care about smooth curvature itself
+- A `solver=` option, letting users pick the linear algebra backend for the fitting problem
```
ev-br, line 354: Nit: pick the minimizer

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+- Tutorial for `scipy.interpolate.make_smoothing_spline`
+
+Here is the PR: [scipy/scipy#26168](https://github.com/scipy/scipy/pull/26168)
+
+## Future Work
+
+SciPy still needs work on some features users may want, and I'm planning to work on them after the Quansight Internship. Here are a few likely candidates:
+
+- P-spline penalties, a cheaper discrete alternative to the exact integral penalty
+- Higher order derivative penalties, penalize the third derivative instead of the second when you care about smooth curvature itself
```
ev-br: Nit: generalize  to higher-order splines and (so far, `make_smoothing_spline`  only constructs a cubic spline with the second-derivative penalty term)

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+
+- Tutorial for `scipy.interpolate.make_smoothing_spline`
+
+Here is the PR: [scipy/scipy#26168](https://github.com/scipy/scipy/pull/26168)
+
+## Future Work
+
+SciPy still needs work on some features users may want, and I'm planning to work on them after the Quansight Internship. Here are a few likely candidates:
+
+- P-spline penalties, a cheaper discrete alternative to the exact integral penalty
```
ev-br: Nit: consider stressing that it's a popular alternative to smoothing splines (may or may not give a reference to Marx&Eilers three-pager, https://psplines.bitbucket.io/Support/WhyPsplines.pdf)

apps/labs/posts/fit-a-smooth-curve-using-scipy.mdx
```diff
@@ -0,0 +1,368 @@
...
+- Higher order derivative penalties, penalize the third derivative instead of the second when you care about smooth curvature itself
+- A `solver=` option, letting users pick the linear algebra backend for the fitting problem
+- Wrapping LAPACK's `dpbcon` in `scipy.linalg`, which gives a cheap condition number estimate from the Cholesky factor we already compute
```
ev-br: Estimate the condition number and warn the user if the problem is detected to be ill-conditioned (this requires wrapping a LAPACK `dpbcon` subroutine)

ev-br (commented):
Reads pretty good! 
Several optional comments below.

ev-br:
My comments have all been addressed, modulo two points:  a stable DOI identifier for the companion report (which holds merging the SciPy PR, too), and a minor wording nitpick around the `solver=` future work item.

AFAICS, Melissa's comments are all resolved, too. What about your comments @czgdp1807 , are you happy with the current state?

czgdp1807 (approved):
LGTM. TY again. 

I will review your related SciPy PRs today. In addition to this, I leave it upto you whether to merge this blog post before or after your related SciPy PRs are merged.

aadya940:
The `solver=` point has been fixed now, and it LGTM. For the permanent URL, we’ve decided to wait until the 29th. If arXiv posts it by then, we’ll go with that; otherwise, we’ll use Zenodo (_which is instant_). If we can edit the blog post later, then there’s no need to wait!

Thanks again @ev-br @melissawm @czgdp1807!

melissawm:
We can wait, no problem. Please ping me when you have the other stuff sorted out and I'll merge. Thanks all! ❤️

